PR #19, opened 2023-11-03
worker: remove `allowed-users`

zowoq:
The nix default for `allowed-users` is `*` so either need to remove this or change it to `extra-allowed-users`.

Or is this meant to be `trusted-users`?


Mic92:
This should work without trusted-users, I went for `extra-allowed-users` because this should than work also for users that set `allowed-users`: https://github.com/Mic92/buildbot-nix/pull/21

PR #26, opened 2023-11-06
Drop secret patch

Mic92:
we have to wait until https://github.com/buildbot/buildbot/pull/7159 is part of a new release.


Mic92:
@mergify queue

PR #27, opened 2023-11-06
Also pass GITHUB_TOKEN to AvatarGithub

Mic92:


buildbot_nix/__init__.py
```diff
@@ -719,16 +719,26 @@ def configure(self, config: dict[str, Any]) -> None:
             projects = [p for p in projects if self.github.topic in p.topics]
         worker_config = json.loads(read_secret_file(self.nix_workers_secret_name))
         worker_names = []
-        config["workers"] = config.get("workers", [])
+
+        for key in [
+            "builders",
```
phaer, line 724: Some of them should already be defined https://github.com/buildbot/buildbot/blob/09a54abffc1ea98dae2fe39df058759c2930a118/master/buildbot/configurators/__init__.py#L36

Mic92: also did: https://github.com/buildbot/buildbot/pull/7209/files

PR #28, opened 2023-11-07
fix project name interpolation for manual builds

phaer:
The project property seems only to be set by the webhook, it's undefined on manually triggered builds for me (and @aciceri who reminded me to push this now)


Mic92:
@mergify queue

Mic92:
@mergify unqueue

PR #30, opened 2023-11-09
master: add prometheus metrics

zowoq:
Currently it is a bit awkward to add extra config and packages, for now can we add this here and if it is useful look at a nicer integration and upstreaming the package to nixpkgs?

### DIFF
--- a/nix/master.nix
+++ b/nix/master.nix
@@ -87,6 +87,12 @@ in
         default = null;
         example = "/var/www/buildbot/nix-outputs";
       };
+
+      prometheusExporterPort = lib.mkOption {
+        type = lib.types.nullOr lib.types.port;
+        default = null;
+        description = "Port where metrics will be served";
+      };
     };
   };
   config = lib.mkIf cfg.enable {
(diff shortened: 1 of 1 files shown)


nix/master.nix
```diff
@@ -117,6 +117,7 @@ in
         c["www"]["plugins"].update(
             dict(base_react={}, waterfall_view={}, console_view={}, grid_view={})
         )
+        c['services'].append(reporters.Prometheus(port=8011))
```
Mic92, line 120: The port might collide with something else:

```suggestion
        ${lib.optionalString (cfg.prometheusExporterPort != null) ''
          c['services'].append(reporters.Prometheus(port=${cfg.prometheusExporterPort}))
        ''}
```

nix/master.nix
```diff
@@ -160,6 +161,18 @@ in
         pkgs.buildbot-plugins.wsgi-dashboards
         pkgs.buildbot-plugins.badges
         (pkgs.python3.pkgs.callPackage ../default.nix { })
+        (ps.buildPythonPackage rec {
```
Mic92, line 164: This would be than `++ lib.optional (cfg.prometheusExporterPort != null) ...`

zowoq:
https://github.com/Mic92/buildbot-nix/commit/0b145d91d8d52b52ddad6d341ed649e4a3e1e4bd

PR #34, opened 2023-11-12
Simplify retry, fix github avatars, make evalWorkerCount configurable

Mic92:


buildbot_nix/__init__.py
```diff
@@ -816,7 +822,7 @@ def configure(self, config: dict[str, Any]) -> None:
         if not config["www"].get("auth"):
             config["www"]["avatar_methods"] = config["www"].get("avatar_methods", [])
             config["www"]["avatar_methods"].append(
-                util.AvatarGitHub(self.github.token())
+                util.AvatarGitHub(token=self.github.token())
```
zowoq, line 825: Shouldn't this use the oauth app instead of the token?

Mic92: I was thinking about that but currently, we don't require the client secret yet for anything else, that we would need here. This could be changed once we completely are able to replace the GITHUB_TOKEN by the github app.

Mic92: But for that to happen some smaller upstream patches are required.

zowoq: > This could be changed once we completely are able to replace the GITHUB_TOKEN by the github app.

What do we need to for this?

Mic92: I recall there was at least one instance where it wouldn't use `renderSecrets` to re-load the github token.

Mic92: If it consistently uses that than we can update the github token in a file periodically using the github app credentials. I have the code for that already in nixpkgs-merge-bot.

Mic92: But I cannot find it just now. Maybe it was fixed as well?

Mic92: I am searching for "Authorization" in buildbot for that.

PR #39, opened 2023-11-18
Move skipped builds to local worker

Mic92:


zowoq:
@Mic92 While watching builds being processed it looks like the skipped builds aren't parallelized?

MagicRB:
https://github.com/buildbot/buildbot/blob/c6dc292f38d0353d33215f8809c617afa47e11a6/master/buildbot/worker/base.py#L645 according to here the worker should be able to attach to multiple builders. We don't set `max_builds` our selves. I'll try to experiment with this and see if i can get it parallel.

zowoq:
Thanks.

I remembered that the lix fork did something with the skipped builds, they actually removed it, not sure if that is relevant here or not.

https://git.lix.systems/lix-project/buildbot-nix/commit/156e6e3deae3edd06de7de1dc1dd20de0b6a4c74

Mic92:
The idea of this was to avoid having to do network I/O or take a lock for skipped builds. But I never properly tested how much of a performance difference this makes.

PR #80, opened 2024-01-08
only refresh github projects in timer and sort them

Mic92:
This way errors are more visible to users on the first running.


Mic92:
@mergify queue

Mic92:
@mergify unqueue

Mic92:
@mergify queue

PR #90, opened 2024-01-30
nix: enable buildbot badges

zimbatm:
Just enough config to make this easily extensible in the future.


zimbatm:
It needs more work. When deploying this, the process starts fine, but nginx still return 500 errors.

nix/master.nix
```diff
@@ -183,9 +192,19 @@ in
...
         ps.psycopg2
         (ps.toPythonModule pkgs.buildbot-worker)
+        pkgs.buildbot-plugins.badges
         pkgs.buildbot-plugins.www-react
         (pkgs.python3.pkgs.callPackage ../default.nix { })
       ];
+
+      extraConfig = ''
+        # WOOT
+        c['www'] = json.loads(
```
Mic92, line 202: I think you are overwriting existing www settings here.

```
c['www'].append(json.loads(${builtins.toJSON www_config}))

Comment (Mic92) [reply nix/master.nix:202]: the nixos module does this:

```
  defaultMasterCfg = pkgs.writeText "master.cfg" ''
    from buildbot.plugins import *
    ${cfg.extraImports}
    factory = util.BuildFactory()
    c = BuildmasterConfig = dict(
     workers       = [${concatStringsSep "," cfg.workers}],
     protocols     = { 'pb': {'port': ${toString cfg.pbPort} } },
     title         = '${escapeStr cfg.title}',
     titleURL      = '${escapeStr cfg.titleUrl}',
     buildbotURL   = '${escapeStr cfg.buildbotUrl}',
     db            = dict(db_url='${escapeStr cfg.dbUrl}'),
     www           = dict(port=${toString cfg.port}),
     change_source = [ ${concatStringsSep "," cfg.changeSource} ],
     schedulers    = [ ${concatStringsSep "," cfg.schedulers} ],
     builders      = [ ${concatStringsSep "," cfg.builders} ],
     services      = [ ${concatStringsSep "," cfg.reporters} ],
     configurators = [ ${concatStringsSep "," cfg.configurators} ],
    )
    for step in [ ${concatStringsSep "," cfg.factorySteps} ]:
      factory.addStep(step)

    ${cfg.extraConfig}
  '';
```

Maybe it should more look like this:

```
  defaultMasterCfg = pkgs.writeText "master.cfg" ''
    from buildbot.plugins import *
    ${cfg.extraImports}
    factory = util.BuildFactory()
    c = BuildmasterConfig = dict(
     workers       = [${concatStringsSep "," cfg.workers}],
     protocols     = { 'pb': {'port': ${toString cfg.pbPort} } },
     title         = '${escapeStr cfg.title}',
     titleURL      = '${escapeStr cfg.titleUrl}',
     buildbotURL   = '${escapeStr cfg.buildbotUrl}',
     db            = dict(db_url='${escapeStr cfg.dbUrl}'),
     www           = dict(port=${toString cfg.port}, **json.load(${builtins.toJSON www_config})),
     change_source = [ ${concatStringsSep "," cfg.changeSource} ],
     schedulers    = [ ${concatStringsSep "," cfg.schedulers} ],
     builders      = [ ${concatStringsSep "," cfg.builders} ],
     services      = [ ${concatStringsSep "," cfg.reporters} ],
     configurators = [ ${concatStringsSep "," cfg.configurators} ],
    )
    for step in [ ${concatStringsSep "," cfg.factorySteps} ]:
      factory.addStep(step)

    ${cfg.extraConfig}
  '';
```

Comment (Mic92) [reply nix/master.nix:202]: Mhm. I though about it and it might be not feasible with just plain json:

i.e you need classes in there:

```python
            config["www"]["auth"] = util.GitHubAuth(
                self.github.oauth_id,
                read_secret_file(self.github.oauth_secret_name),
                apiVersion=4,
            )
```

zimbatm: Taking a step back, it's probably only worth doing something like this either directly upstream, or by merging the upstream module into this one. Otherwise both are fighting to control the base data structures.

We could have "settings" that also supports python code snippets as a value type. It wouldn't be great for programming, but probably OK for small snippets like that.

The other option is to have a pure data-only pass to fill the tree, and then ask the user to complement it with code snippest in a second pass.

zimbatm:
https://github.com/buildbot/buildbot/blob/bb4354043f45be5f32e80deb90896a96fcfa3971/www/badges/buildbot_badges/__init__.py#L84-L94

This plugin may not work for us. Our builder names have slashes in them and I think their router doesn't support that.

Mic92:
This one might help: https://github.com/buildbot/buildbot/pull/7805

Mic92:
Might be fixed with: https://github.com/buildbot/buildbot/pull/7805

zowoq:
I don't think the badge links would be stable, they'd break if the database was reset?

PR #124, opened 2024-04-22
Add support for Gitea

MagicRB:
This pull request is just there to publicize the code I'm working on, before being ready it'll get probably more cleaned up and merged with the Gitea backend pull request.

### DIFF
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -9,17 +9,23 @@ from collections import defaultdict
 from collections.abc import Generator
 from dataclasses import dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Callable
+import typing
 
+from buildbot.process.log import StreamLog
 from buildbot.configurators import ConfiguratorBase
 from buildbot.interfaces import WorkerSetupError
-from buildbot.plugins import reporters, schedulers, secrets, steps, util, worker
+from buildbot.plugins import reporters, schedulers, steps, util, worker
 from buildbot.process import buildstep, logobserver, remotecommand
 from buildbot.process.project import Project
 from buildbot.process.properties import Interpolate, Properties
 from buildbot.process.results import ALL_RESULTS, statusToString
 from buildbot.steps.trigger import Trigger
 from buildbot.www.authz.endpointmatchers import EndpointMatcherBase, Match
+from buildbot.www.authz import Authz
+from buildbot.secrets.providers.file import SecretInAFile
+from buildbot.locks import MasterLock
+from buildbot.config.builder import BuilderConfig
 
 if TYPE_CHECKING:
     from buildbot.process.log import Log
@@ -28,12 +34,23 @@ from twisted.internet import defer, threads
 from twisted.logger import Logger
 from twisted.python.failure import Failure
 
+from .gitea_projects import (
+    GiteaConfig
+)
+
 from .github_projects import (
-    GithubProject,
-    create_project_hook,
-    load_projects,
-    refresh_projects,
+    GithubBackend,
+    GithubConfig,
     slugify_project_name,
+)
+
+from .projects import (
+    GitProject,
+    GitBackend
+)
+
+from .secrets import (
+    read_secret_file
 )
 
 SKIPPED_BUILDER_NAME = "skipped-builds"
@@ -48,8 +65,11 @@ class BuildbotNixError(Exception):
 class BuildTrigger(Trigger):
     """Dynamic trigger that creates a build for every attribute."""
 
+    project: GitProject
+
     def __init__(
         self,
+        project: GitProject,
         builds_scheduler: str,
         skipped_builds_scheduler: str,
         jobs: list[dict[str, Any]],
@@ -57,6 +77,7 @@ class BuildTrigger(Trigger):
     ) -> None:
         if "name" not in kwargs:
             kwargs["name"] = "trigger"
+        self.project = project
         self.jobs = jobs
         self.config = None
         self.builds_scheduler = builds_scheduler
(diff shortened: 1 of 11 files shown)


buildbot_nix/__init__.py
```diff
@@ -239,7 +262,9 @@ def __init__(self, **kwargs: Any) -> None:
     @defer.inlineCallbacks
     def run(self) -> Generator[Any, object, Any]:
         # run `nix build`
-        cmd: remotecommand.RemoteCommand = yield self.makeRemoteShellCommand()
```
Mic92, line 242: What was the issue with the previous line?

buildbot_nix/__init__.py
```diff
@@ -224,7 +245,9 @@ def run(self) -> Generator[Any, object, Any]:
         error = self.getProperty("error")
         attr = self.getProperty("attr")
         # show eval error
-        error_log: Log = yield self.addLog("nix_error")
+        # TODO why doesn't type information pass through yield again?
+        error_log_: object  = yield self.addLog("nix_error")
```
Mic92, line 249: same question as above.

buildbot_nix/__init__.py
```diff
@@ -145,17 +164,22 @@ class NixEvalCommand(buildstep.ShellMixin, steps.BuildStep):
...
         super().__init__(**kwargs)
+        self.project = project
         self.observer = logobserver.BufferLogObserver()
         self.addLogObserver("stdio", self.observer)
         self.supported_systems = supported_systems
 
     @defer.inlineCallbacks
     def run(self) -> Generator[Any, object, Any]:
         # run nix-eval-jobs --flake .#checks to generate the dict of stages
-        cmd: remotecommand.RemoteCommand = yield self.makeRemoteShellCommand()
```
Mic92, line 158: Same question as above.

buildbot_nix/__init__.py
MagicRB: All of your questions, types, typechecking seemed to be broken for me, soemhow through a yield it wouldnt type check, my guess is that a yield can return an error value. I wanted to remove all the type errors so its easier for me to write (and for the future). I can revert these bits no problem if so desired

buildbot_nix/__init__.py
```diff
@@ -347,8 +322,9 @@ def run_vc(
             return res
 
         # The code below is a modified version of Git.run_vc
-        self.stdio_log: Log = yield self.addLogForRemoteCommands("stdio")
-
+        # TODO why doesn't type information pass through yield again?
+        stdio_log_: object = yield self.addLogForRemoteCommands("stdio")
```
Mic92, line 326: Same question as above.

buildbot_nix/__init__.py
Mic92: What typechecker are you using?

MagicRB: pyright

MagicRB: I will try mypy then and clean up these workarounds 👍. Sorry for the mess

Mic92: Alright. You can use `nix fmt` to get a local type check.

MagicRB:
Ready for review! I verified all functionality to be working against a fork of this repository at [https://github.com/magicrb/buildbot-nix](https://github.com/magicrb/buildbot-nix) and [https://codeberg.org/magic_rb/buildbot-nix](https://codeberg.org/magic_rb/buildbot-nix). No regressions have been introduced as far as I can tell. The following section includes notes that I took while working on this.

# Initial Setup

-   confused about how the GH integration works. The docs aren't great.
-   <https://github.com/settings/tokens/new> took me ages to find
-   the gh app must be installed and oauth enabled
-   git is required
-   force eval manually is broken, properties do not get populated
-   will only consider PRs or default branch, no way to make it consider other branches

# Removing GitHub Assumption

-   gonna end up using <https://github.com/lab132/buildbot-gitea>, thank god I don't have to implement this myself;
-   <https://github.com/microsoft/pyright/issues/6000> so therefore `buildbot.plugins.util.BuilderConfig` is invalid?
    -   `from buildbot.config.builder import BuilderConfig` works

# Adding Gitea

-   Gitea's API is **very** similar to GitHub's, especially the responses. Only thing different so far is that Gitea doesn't include `topics` in the repo listing, but that is easy to splice in.
-   It is very unclear whether Gitea includes repos of orgs a user belongs to in the `/api/v1/user/repos` endpoint, if not it's easy to splice in again
-   what happens if the same project `owner/repo` is on Gitea and GitHub
-   the reporters are stupid, they will trigger all for every repo, i dont know how to make them not trigger for the incorrect one, maybe subclassing and a filter?s

# Cleanup and Finalizing

-   still don't know how to resolve this

    ```python
    # TODO Gitea doesn't include this information
    return False # self.data["owner"]["type"] == "Organization"
    ```

    Gitea really doesn't seem to expose that information, [the API docs](https://forgejo.dev/api/swagger#/repository/repoGet) agree.


Lastly:
mypy is a much weaker type checker in my experience, it would miss many cases of errors that pyright would report, but I did undo the weird type casting since we're using mypy.

mannp:
Very happy for this, thank you. Out of interest, would this support private forgejo/gitea repos or public only?

MagicRB:
> Very happy for this, thank you. Out of interest, would this support private forgejo/gitea repos or public only?

If the buildbot instance can access the instance, https works, then it should be fine I think. It is untested though.

MagicRB:
oh yeah, there are a few unresolved things: 
1. reporters will trigger for both gitea and github, both of them. The gitea one will skip the report gracefully, the github one will throw an exception. But it seems to be harmless
2. The Gitea plugin doesnt support avatars
3. Gitea doesn't seem to distinguish between organizations and users in its API, so `belongs_to_org` is a noop and always returns `False`
4. buildbot doesn't support multiple authentication plugins at once, the default is `github`

buildbot_nix/github_projects.py
```diff
@@ -1,93 +1,229 @@
...
+            )
+            return util.FAILURE
+
+
+@dataclass
+class GithubConfig:
+    oauth_id: str
+
+    # TODO unused
+    buildbot_user: str
```
zowoq, line 71: Does `TODO unused` mean `buildbot_user` can be removed?

MagicRB: At the time I wasn't a 100% sure, but now I am, and yes it can

zowoq: Thanks, I'll send a PR.

PR #146, opened 2024-05-08
Make the webhook endpoint URL configurable

antifuchs:
In some CI setups (especially those where the CI server is reachable over a VPN with split-horizon DNS), the URL to access the CI server differs from the URL that the (external) webhook endpoint listens on. 

This change introduces an option webhookBaseUrl that allows setting that URL independently from the buildbot installation URL.

### DIFF
--- a/nix/master.nix
+++ b/nix/master.nix
@@ -181,6 +181,13 @@ in
         example = "buildbot.numtide.com";
       };
 
+      webhookBaseUrl = lib.mkOption {
+        type = lib.types.str;
+        description = "URL base for the webhook endpoint that will be registered for github or gitea repos.";
+        example = "https://buildbot-webhooks.numtide.com/";
+        default = "${config.services.buildbot-master.buildbotUrl}";
+      };
+
       outputsPath = lib.mkOption {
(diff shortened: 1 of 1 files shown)


nix/master.nix
```diff
@@ -181,6 +181,13 @@ in
         example = "buildbot.numtide.com";
       };
 
+      webhookBaseUrl = lib.mkOption {
+        type = lib.types.str;
+        description = "URL base for the webhook endpoint that will be registered for github or gitea repos.";
+        example = "https://buildbot-webhooks.numtide.com/";
+        default = "${config.services.buildbot-master.buildbotUrl}";
```
Mic92, line 188:
```suggestion
        default = config.services.buildbot-master.buildbotUrl;
```

amnnku:
Out of interest, does setting the webhook url manually change the need for the user to be an admin in gitea? 

Or do you still set the url automatically, just with this revised url.

antifuchs:
You still have to have permissions to set the webhook endpoint on each repo. This only alters the URL that the  buildbot-nix code sets using those permissions.

Mic92:
@mergify queue

PR #148, opened 2024-05-09
master: Add config to force https for the endpoint

Erethon:
Let me preface this by saying I'm new to Nix, so sorry if I've missed something obvious and there's another way around this. Also, thank you for your work on this module!

I'm setting up  a buildbot instance using buildbot-nix and I ran into a usecase that isn't currently handled by the module. My buildbot is behind a third party proxy that does TLS termination for me, so I don't need nginx to create or serve any certificates, but there's no way to let the module know that the endpoint must start with `https://` because this is currently tied to the nginx configuration. This patch adds a config option to make the endpoint use `https://` even if it's not setup in nginx.

### DIFF
--- a/examples/master.nix
+++ b/examples/master.nix
@@ -69,4 +69,8 @@
   #  forceSSL = true;
   #  enableACME = true;
   #};
+
+  # Optional: If buildbot is setup to run behind another proxy that does TLS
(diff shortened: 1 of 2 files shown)


Mic92:
@mergify queue

PR #149, opened 2024-05-09
master: on-failure -> always

zowoq:
https://github.com/nix-community/infra/pull/1235#discussion_r1590512432

Seems this needs to be always instead of on-failure as the service exits successfully.


Mic92:
@mergify queue

Issue #173, opened 2024-06-11
Build the frontend from source


zimbatm:
Right now buildbot's npm frontend mess is getting copied directly from the release. This makes it harder for us to experiment with UI changes.

PR #180, opened 2024-06-12
Parse evaluation warnings and display in separate section.

### LINKED ISSUE #172: Parse the Nix evaluation warnings and display them in the UI
It would be nice to display warnings so they stand out in the UI.

Right now if the user doesn't look at the logs, they might miss deprecation issues for example.

MagicRB:
A live version can be seen [here](https://buildbot.redalder.org/#/builders/7/builds/52) and in case my buildbot dies or gets reset, a screenshot:

image

Possibly fixes #172, for anything more fancy we'll need #173, what do you think @zimbatm?

And due to the exploration here, I've learned that we can add arbitrary HTML to the page, as long as it is contained in one of those tabs. Well, injecting javascript and modifying the page is an option, but a really bad one :)

### DIFF
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -159,7 +159,7 @@ class NixEvalCommand(buildstep.ShellMixin, steps.BuildStep):
         kwargs = self.setupShellMixin(kwargs)
         super().__init__(**kwargs)
         self.project = project
-        self.observer = logobserver.BufferLogObserver()
+        self.observer = logobserver.BufferLogObserver(wantStderr=True)
         self.addLogObserver("stdio", self.observer)
         self.supported_systems = supported_systems
 
@@ -172,6 +172,11 @@ class NixEvalCommand(buildstep.ShellMixin, steps.BuildStep):
         # if the command passes extract the list of stages
         result = cmd.results()
         if result == util.SUCCESS:
+            log.info(self.observer.getStderr())
+            self.addHTMLLog(
+                "Evaluation Warnings", f"<pre>{self.observer.getStderr()}</pre>"
+            )
+
             # create a ShellCommand for each stage and add them to the build
             jobs = []
 


buildbot_nix/__init__.py
```diff
@@ -172,6 +172,11 @@ def run(self) -> Generator[Any, object, Any]:
         # if the command passes extract the list of stages
         result = cmd.results()
         if result == util.SUCCESS:
+            log.info(self.observer.getStderr())
+            self.addHTMLLog(
+                "Evaluation Warnings", f"<pre>{self.observer.getStderr()}</pre>"
```
zimbatm, line 177: Do you know if the string interpolation gets properly HTML-escaped?

buildbot_nix/__init__.py
```diff
@@ -172,6 +172,11 @@ def run(self) -> Generator[Any, object, Any]:
         # if the command passes extract the list of stages
         result = cmd.results()
         if result == util.SUCCESS:
+            log.info(self.observer.getStderr())
```
zimbatm, line 175: Could you make the eval errors more visible? As a user, I would see this when looking at the build:

image

Then click on two toggles to unveil the eval error:

image

(small nit: the whitespace in front of the first line)

zimbatm (commented):
nice, this is on the right track

buildbot_nix/__init__.py
MagicRB: Whitespace is already fixed, lefr over from a test and me not knowing how `pre` works. But i dont think we can autoexpand, ill look. But to not put it therw would need UI changes

buildbot_nix/__init__.py
MagicRB: Uh, definitely not, good catch! JavaScript injection attacks disaprove :)

MagicRB:
Anything more than this is blocked by #173 and further discussion is needed, as buildbot does not allow partial UI replacements as far as I can tell, therefore we cannot add another element into the build UI.

buildbot_nix/__init__.py
Mic92: I think the python stdlib has some helpers for that. Otherwise I am sure that buildbot has it somewhere.

MagicRB: Indeed, but this was more of a prototype to showcase what can be done now anyway, but i don't think it's worth merging as it is, personally.

buildbot_nix/__init__.py
zimbatm: What about the 
image
Could this be turned orange somehow?

zimbatm:
Unless the eval error gets more visibility, it makes sense to table this for now.

buildbot_nix/__init__.py
MagicRB: I will check, but I dont have high hopes

MagicRB: Builbots UI is very configurable, if you want to add/replace views, not if you want to modify them...

Mic92:
@mergify rebase

MagicRB:
This is superseded by #255 as now evaluation failures appear as separate builds. 
image

Mic92:
But is this the same as warnings though?

Mic92:
> But is this the same as warnings though? Those

MagicRB:
Ah yeah if the build doesn't fail it doesn't show. That's my mistake.

PR #182, opened 2024-06-12
Defer webhook creation to project reload

MagicRB:


MagicRB:
Do *NOT* test now, will end up creating a webhook for every single repo it has access to :/

MagicRB:
Will keep on my instance for a while and check over once more, this is a way more intrusive change than I was hoping for.

Mic92:
@mergify unqueue

Mic92:
@mergify rebase

Mic92:
I actually just wanted to rebase this, not merge it.

MagicRB:
Did we merge this by another PR? The code from this seems to be in `master`

zowoq:
> Did we merge this by another PR? The code from this seems to be in `master`

Looks like https://github.com/nix-community/buildbot-nix/pull/203?

MagicRB:
Oh yeah, you're right, i even wrote that code :sweat_smile: I really have the memory of a goldfish (might be an insult to goldfish all over the world)

PR #194, opened 2024-06-25
drop pre 24.05 patch

zowoq:
4.0 seems to retain compatibility with 3.11 so I'll leave it as is for now and just drop the outdated patch.


Mic92:
@mergify queue

PR #198, opened 2024-06-25
Add local worker support

james-atkins:
If `pass` does not exist, then create a local worker that runs in the same process as the master. 

https://docs.buildbot.net/latest/manual/configuration/workers.html#local-workers


Mic92:
What was the issue with this approach? In general I find it nicer to have the two components distributed over two services. Because oom events happen in eval will only crash workers but the webui will still be able to report.

PR #200, opened 2024-06-27
Fix typo, secretKeyFile is a path not a string

Erethon:
Minor documentation fix after the recent GitHub App changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ To integrate with GitHub using app authentication:
    - Set
      `services.buildbot-nix.master.github.authType.app.id = <your-github-id>;`
    - Set
-     `services.buildbot-nix.master.github.authType.app.secretKeyFile = "/path/to.pem";`
+     `services.buildbot-nix.master.github.authType.app.secretKeyFile = /path/to.pem;`
 3. **Install App**: Install the app for an organization or specific user.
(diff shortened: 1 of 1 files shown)


Mic92:
Does it not accept strings at all? We actually want that to not import secrets into the nix store.

MagicRB:
It ahould not be a path, if anything we should deny that in the module.

Mic92:
What does your string looks like? This is what I have set:

```
nix-repl> nixosConfigurations.eve.config.services.buildbot-nix.master.github.authType.app.secretKeyFile
"/run/secrets/buildbot-github-app-secret-key"
```

It's by the way the same as for all other secret option. The string needs to contain an absolute path.

Erethon:
> What does your string looks like? This is what I have set:
> 
> ```
> nix-repl> nixosConfigurations.eve.config.services.buildbot-nix.master.github.authType.app.secretKeyFile
> "/run/secrets/buildbot-github-app-secret-key"
> ```
> 
> It's by the way the same as for all other secret option. The string needs to contain an absolute path.

I didn't realize the string had to be an absolute path and I was using a relative one. Ehen I saw the code and error message it made sense to me to have this as a path and not a string. Closing this as invalid then, thank you!

PS. Sorry for the late reply, I was afk for a while.

PR #203, opened 2024-07-03
Defer hook creation to project reload, both GitHub and Gitea

MagicRB:
Defers hook creation to project reload time, adds another `BuildStep` to each reload `buildset` which creates hooks for all loaded projects. As a side effect, filters repositories by tag, before writing them to the cache file, keeps load-time filter to not break existing deployments.


MagicRB:
Verified to work with Gitea, GitHub App, and GitHub Legacy

Mic92:
@mergify rebase

PR #209, opened 2024-07-09
remove automatic retry

zowoq:
Automatic retry is nice when the build is cheap, not so much if it is a multi-hour build.

I don't see a problem with removing this now that now that everyone has restart permissions.

### DIFF
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -2,7 +2,6 @@ import json
 import multiprocessing
 import os
 import re
-import uuid
 from collections import defaultdict
 from collections.abc import Generator
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


Mic92:
nix-community is a bit of a special case because we have many users there, however for many smaller instances the default should be ok. I made this now configurable: https://github.com/nix-community/buildbot-nix/pull/210

zowoq:
Thanks.

zowoq:
How do I set zero retries?

```
      buildRetries = lib.mkOption {
        type = lib.types.ints.positive;
        default = 1;
        description = "Number of times a build is retried";
      };
```

Mic92:
Does ints.positive doesn't allow 0? we can also change it to int otherwise.

zowoq:
```
error: A definition for option `services.buildbot-nix.master.buildRetries' is not of type `positive integer, meaning >0'
```

Mic92:
https://github.com/nix-community/buildbot-nix/pull/211

zowoq:
https://buildbot.nix-community.org/#/builders/37/builds/70

Just noticed this, got caught in a loop somehow?  Looked like it was happening in a couple of different repos.

I've reverted it for now. https://github.com/nix-community/infra/pull/1365

PR #215, opened 2024-07-11
Don't pass token into `AvatarGitHub`, it works perfectly without it

MagicRB:
This removes the runtime modifying of the `AvatarGitHub` class.

After closer inspection of the code and the GitHub API [docs](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-a-user) and [these](https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-a-user) I realized, both these endpoints work fine without a token. The issue might be that the ratelimits are stricter without authentication, but we're planning to implement retries anyway, so that shouldn't be an issue for now.


Mic92:
@mergify rebase

Mic92:
This was merged in https://github.com/nix-community/buildbot-nix/pull/218

PR #216, opened 2024-07-11
Support per installation tokens directly in `GitHubStatusPush`

MagicRB:
This PR creates a patch for buildbot supporting a `token` argument being a function, which given a `reports` object, gives back which token to use. It caches this so it doesn't have to re-ask for it and call `render` on each request. 

PR is a draft for now as it's build for buildbot 3, I first need to bump my instance and then rebase the patch, we should probably keep both patches for 3 and 4.


Mic92:
Yes. Stable NixOS has buildbot 3 and it's usually a painpoint to backport. However if our patches grow bigger over the summer, it might be better to vendor buildbot to reduce maintenance effort.

zimbatm:
> However if our patches grow bigger over the summer, it might be better to vendor buildbot to reduce maintenance effort.

And also the buildbot NixOS module. I noticed that reasoning becomes complicated with both level of modules. Since it's an opinionated distribution of buildbot, we don't need all that flexibility.

Mic92:
@mergify rebase

Mic92:
I made some change how the vendoring works:

```
The overlay will break if buildbot in unstable will change it's python
override in nixpkgs.
This will happen in the next release when sqlalchemy will be update.

Also I have at least three installations where I share nixpkgs instances
across multiple nixos configuration for performance reasons,
where I cannot use overlays defined in a NixOS module.
```

nix/0001-Support-per-installation-tokens-in-GithubStatusPush-.patch
```diff
@@ -0,0 +1,66 @@
...
+--- a/master/buildbot/reporters/github.py
++++ b/master/buildbot/reporters/github.py
+@@ -70,7 +70,14 @@ class GitHubStatusPush(ReporterBase):
+         generators=None,
+         **kwargs,
+     ):
+-        token = yield self.renderSecrets(token)
++        headers = {}
++        if not callable(token):
++            token = yield self.renderSecrets(token)
```
Mic92, line 23: I think the renderSecrets interface would be an alternative approach to this actually. One can also define an object that would the same thing we are currently doing.

MagicRB:
@Mic92 should I pull the commuts you made into #219 ?

Mic92:
> @Mic92 should I pull the commits you made into #219 ?

I can just merge together with the actual feature.

nix/0001-Support-per-installation-tokens-in-GithubStatusPush-.patch
Mic92: Mhm. But than it's not per installation...

Mic92:
^ will test this approach now. It will make the patch even simpler.

MagicRB:
Huh, did not know that was a thing, good to know, how did you find that?

Mic92:
I knew about the IRenderable interface. Than I was just looking for a good candidate that implemented it.

MagicRB:
Ah okay, that didnt occur to me, good catch

MagicRB:
I hope the rendering isn't slow

Mic92:
> I hope the rendering isn't slow

It's the same function in the end that also renders the githhub status name.

Mic92:
Depending on what solution buildbot accepts upstream, we can switch than our implementation around. But let's go with this.

Mic92:
@mergify queue

PR #217, opened 2024-07-11
Combined build reports GitHub

MagicRB:
This PR adds two derived classes `ModifyingGitHubStatusPush` and `ModifyingGiteaStatusPush`, which take the `modifyingFilter` argument. It has the type `Callable[[Any], Any | None]`, if it returns `None` the report is canceled, if it returns non `None` the returned object is used instead of the original.

We use this addition to implement "combined builds", if the jobs count for a `nix-eval` is over a certain number (made to be configurable in a future rendition of this PR) `buildbot-nix` will not create reports on GitHub for each build, but only a single `nix-build-combined` report which indicates whether all builds succeeded or failed.

~~This PR is also a draft because it is built for buildbot 3, as with #216 I first need to bump my instance to 4. I'll make a separate patch for 3 and 4 so we keep compatibility.~~

### TODO 
- ~implement for Gitea~
- ~make the limit configurable.~
- `nix-build-combined` appears only after all builds complete

### DIFF
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -165,7 +165,11 @@ class NixEvalCommand(buildstep.ShellMixin, steps.BuildStep):
     project: GitProject
 
     def __init__(
-        self, project: GitProject, supported_systems: list[str], **kwargs: Any
+        self,
+        project: GitProject,
+        supported_systems: list[str],
+        job_report_limit: int | None,
+        **kwargs: Any,
     ) -> None:
         kwargs = self.setupShellMixin(kwargs)
         super().__init__(**kwargs)
@@ -173,6 +177,7 @@ class NixEvalCommand(buildstep.ShellMixin, steps.BuildStep):
         self.observer = logobserver.BufferLogObserver()
         self.addLogObserver("stdio", self.observer)
         self.supported_systems = supported_systems
+        self.job_report_limit = job_report_limit
 
     @defer.inlineCallbacks
     def run(self) -> Generator[Any, object, Any]:
@@ -205,29 +210,35 @@ class NixEvalCommand(buildstep.ShellMixin, steps.BuildStep):
             self.number_of_jobs = len(filtered_jobs)
 
             self.build.addStepsAfterCurrentStep(
-                [  # noqa: RUF005
+                [
                     BuildTrigger(
                         self.project,
                         builds_scheduler=f"{project_id}-nix-build",
                         skipped_builds_scheduler=f"{project_id}-nix-skipped-build",
                         name="build flake",
                         jobs=filtered_jobs,
-                        report_status=(self.number_of_jobs <= 2),
+                        report_status=(
+                            self.job_report_limit is None
+                            or self.number_of_jobs <= self.job_report_limit
+                        ),
                     ),
                 ]
-                + [
-                    Trigger(
-                        waitForFinish=True,
-                        schedulerNames=[f"{project_id}-nix-build-combined"],
-                        haltOnFailure=True,
-                        flunkOnFailure=True,
-                        sourceStamps=[],
-                        alwaysUseLatest=False,
-                        updateSourceStamp=False,
-                    ),
-                ]
-                if self.number_of_jobs > 2
-                else [],
+                + (
+                    [
+                        Trigger(
+                            waitForFinish=True,
+                            schedulerNames=[f"{project_id}-nix-build-combined"],
+                            haltOnFailure=True,
+                            flunkOnFailure=True,
+                            sourceStamps=[],
+                            alwaysUseLatest=False,
+                            updateSourceStamp=False,
+                        ),
+                    ]
+                    if self.job_report_limit is not None
+                    and self.number_of_jobs > self.job_report_limit
+                    else []
+                ),
             )
 
         return result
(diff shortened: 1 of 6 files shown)


buildbot_nix/__init__.py
```diff
@@ -199,8 +211,19 @@ def run(self) -> Generator[Any, object, Any]:
                         skipped_builds_scheduler=f"{project_id}-nix-skipped-build",
                         name="build flake",
                         jobs=filtered_jobs,
+                        report_status=(self.number_of_jobs <= 2)
                     ),
-                ],
+                ] + [
+                    Trigger(
+                        waitForFinish=True,
+                        schedulerNames=[f"{project_id}-nix-build-combined",],
```
Mic92, line 219: That's cool. Is there a chance to link to the all failed builds in this one as well? I got the feedback that when there long lists of builds it becomes hard to see the failed ones.

MagicRB:
I have a question what default value for the build report combining do we want? I set it to 50 but that may be too high @Mic92

As for the "nix-build-combined appears only after all builds complete" that's a minor thing and I'm not sure how to fix it, I didn't see a way to wait for another buildstep anywhere.

I'll do Gitea support next, still before we merge this as having it implemented for GitHub only would mean that on Gitea both the individual builds and `nix-build-combined` would show up.

Mic92:
Giteas link preview currently only 6 items if you click on the green checkmark for me: https://git.clan.lol/clan/clan-core

Maybe we have to set it quiet low than.

MagicRB:
Well if Gitea limits it to 6, then thats fine no? The limitting on our side is to not overwhelm Gitea/GitHub

Mic92:
I think so.

MagicRB:
I think 50 is reasonable, though maybe 32 just to make it nice.

Mic92:
Big rebase needed here.

MagicRB:
I'll rebase

MagicRB:
Done, checking whether nothing broke.

nix/master.nix
```diff
@@ -266,6 +266,17 @@ in
         default = null;
         example = "/var/www/buildbot/nix-outputs";
       };
+
+      jobReportLimit = lib.mkOption {
+        type = lib.types.nullOr lib.types.ints.positive;
```
zowoq, line 271: > report builds individually as long as the number of builds is less than or equal to `n`

Is zero an option here to only use `nix-build-combined` and never report an individual build?

MagicRB: Setting it to `0`  works, I'll push it after finishing the rest of the tests

MagicRB:
I'm happy with this, ready for review and merging

nix/master.nix
zowoq: Thanks.

zowoq:
Needs to be rebased again. Apart from that, is this ready to merge? Would be useful for nixvim.

Mic92:
It will be probably finished by this week.

MagicRB:
I am unexpectedly not home, forgot we're visiting my grandparents over the weekend, I'm back tuesday evening so yes, this week is the probable finish time.

MagicRB:
Rebased and functional, I've fixed a minor omission from the #225 PR, which may have been there back then too. not sure. Build reports of the form `buildbot/` would pop up, they were the gcroot steps.

It would also be good for me to read up on what `hideStepIf` precisely does. I'll open an issue for what I've noticed.

MagicRB:
I've tested this by resetting all the caches and performing a Gitea and GitHub build.

Mic92:
It's not yet optimal because the build finished even is only printed after all builds, but we want to improve on this. However for now this should unblock nixvim.

MattSturgeon:
> It's not yet optimal because the build finished even is only printed after all builds, but we want to improve on this.

I haven't looked in detail, so maybe this is already discussed/implemented, but some "bigger" checks will print a markdown summary in the check's details page (e.g. mergify's [queue checks](https://github.com/nix-community/nixvim/pull/1933/checks?check_run_id=27938204893))

> However for now this should unblock nixvim.

:heart:

MagicRB:
> I haven't looked in detail, so maybe this is already discussed/implemented, but some "bigger" checks will print a markdown summary in the check's details page (e.g. mergify's [queue checks](https://github.com/nix-community/nixvim/pull/1933/checks?check_run_id=27938204893))
I'm not sure what precisely even is that, I think that's reserved for workflows? Which `buildbot-nix` isn't, but I'll attempt to figure it out.

MagicRB:
According to GitHub [docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks):

> Note: The Checks tab only gets populated for pull requests if you set up checks, not commit statuses, for the repository.

MattSturgeon:
Indeed:

> Checks are different from commit statuses in that they provide line annotations, more detailed messaging, and are only available for use with GitHub Apps.

> [!note]
> GitHub Actions generates checks, not commit statuses, when workflows are run.

> I think that's reserved for workflows?

Workflows (and actions) generate checks, but so can apps (like buildbot and mergify).

I guess currently you're using the statuses API, hence why buildbot doesn't appear in the "checks" tab

MagicRB:
We indeed are, I'll open an issue for this later when I'm back on my computer.

Mic92:
Yes there is an api to append logs to github. Ofborg and garnix are also implementing this. It's not even that much code.

MagicRB:
We just create another reporter class, future PR

MattSturgeon:
Opened as #236

PR #229, opened 2024-07-18
Generalize Cachix support to `post_build_steps`

MagicRB:
This PR generalizes the cachix post build command into a `postBuildCommands` option which allows to run arbitrary commands after `nix build` under the `buildbot-worker`. No sandboxing takes place as these commands are managed by the instance admin.

One thing weird about this code is that I've introduced a custom type called `interpolateType` which has the form `{ _type = "interpolate"; value = "something-%(prop:out_path)s"; }` and allows for arbitrary substitutions to take place. I'm not sure if this weirdness is desired, but in my opinion the power this gives to instance admins is worth the complexity. If this is deemed good I will expand on the documentation to make sure it is not confusing. or at least the littlest amount of confusing humanly possible.


nix/master.nix
```diff
@@ -80,6 +99,60 @@ in
...
+              '';
+            };
+          };
+        });
+
+        example = lib.literalExpression ''
+          [
+            name = "upload-to-s3";
+            environment = {
+              S3_TOKEN = "xxxxxxx";
```
Mic92, line 148:
```suggestion
              S3_TOKEN = "%(secret:s3-token)";
```

PR #231, opened 2024-07-19
Utilize `pydantic` for serialization and deserialization

MagicRB:
Tested with GitHub apps and Gitea, fully working, also closes #230

PR #235, opened 2024-07-26
Add check, which catchis silent disabling of cachix

MagicRB:
Catches a breaking change that silently disables Cachix on existing deployments. @Mic92 please merge ASAP, should have went in with #231 but I thought of this 5 minutes ago.

### DIFF
--- a/nix/master.nix
+++ b/nix/master.nix
@@ -1,6 +1,7 @@
 { config
 , pkgs
 , lib
+, options
 , ...
 }:
 let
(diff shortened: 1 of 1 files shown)


zowoq:
https://buildbot.nix-community.org/#/builders/81/builds/18

This seems to be wrong, don't need both of auth token and signing key.

MagicRB:
Will form a correction then, I've no clue how cachix works, well at least we don:t beak users silently

Mic92:
https://github.com/nix-community/buildbot-nix/pull/237

zowoq:
https://buildbot.nix-community.org/#/builders/81/builds/23

Still seeing the same error.

PR #239, opened 2024-07-26
speed up skipped build processing

Mic92:


MagicRB:
Shouldnt workers be able to attach to multiple builds?

Mic92:
I don't think so. Before I added one worker per CPU, it would only process one build at the time.

MagicRB:
Then what's [that](https://github.com/buildbot/buildbot/blob/c6dc292f38d0353d33215f8809c617afa47e11a6/master/buildbot/worker/base.py#L67)

Mic92:
I checked now a couple of times and with this change the time reduces from 20s to 15s for "skipping" disko

zowoq:
I tried this branch on a fully cached nixvim eval a few days ago. There wasn't any other buildbot/hydra jobs running at the time.

With this branch: `1:06` https://buildbot.nix-community.org/#/builders/26/builds/20

Without: `28:23` https://buildbot.nix-community.org/#/builders/26/builds/21

I ran another one without today: `38:14` https://buildbot.nix-community.org/#/builders/26/builds/95

Mic92:
That's quite significant. Before I had skipped builds not actually having to run any code i.e. build steps. I think they were faster back than.

MagicRB:
https://github.com/buildbot/buildbot/issues/7125 okay this is this, it was not obvious from the code, but yeah, its one builder one worker so it won't run multiple. We should do the same for the `NixEval` builder then

MagicRB:
Oh never mind, we have this in `main` already, or a version of it.

PR #240, opened 2024-07-27
Actually fix Cachix, attempt 3

MagicRB:
This is getting slightly silly, here's me hoping this is the last "Cachix fix" PR...

To that end, I used the following truth table to generate the assertion expression:
[link](https://tma.main.jp/logic/logic.php?lang=en&type=3&v0=a&v1=b&v2=c&00=1&01=1&02=0&03=1&04=0&05=1&06=0&07=1)
image

Where `a`, `b` are the expressions `cfg.cachix.authTokenFile == null` and `cfg.cachix.signingKeyFile == null`, their ordering is meaningless. `c` is `cfg.cachix.enable`. Which leads us to the expression:
```nix
isNull cfg.cachix.authTokenFile && isNull cfg.cachix.signingKeyFile ||
isNull cfg.cachix.authTokenFile && cfg.cachix.enable ||
isNull cfg.cachix.signingKeyFile && cfg.cachix.enable;
```
This catches cases where the user has Cachix enabled but doesn't set `cfg.cachix.enable` to `true`. I've done extensive testing, here are the tested configs and the results:
### 1
```nix
{
    # enable = true;
    name = "foo";
    # auth.signingKey.file = "/file";
    # auth.authToken.file = "/file";
    # signingKeyFile = "/file";
    # authTokenFile = "/file";
}
```
succeeds

### 2 
```nix
{
    # enable = true;
    name = "foo";
    # auth.signingKey.file = "/file";
    # auth.authToken.file = "/file";
    signingKeyFile = "/file";
    # authTokenFile = "/file";
}
```
fails

### 3
```nix
{
    # enable = true;
    name = "foo";
    # auth.signingKey.file = "/file";
    # auth.authToken.file = "/file";
    # signingKeyFile = "/file";
    authTokenFile = "/file";
}
```
fails

### 4
```nix
{
    # enable = true;
    name = "foo";
    # auth.signingKey.file = "/file";
    # auth.authToken.file = "/file";
    signingKeyFile = "/file";
    authTokenFile = "/file";
}
```
fails

### 5
```nix
{
    enable = true;
    name = "foo";
    # auth.signingKey.file = "/file";
    # auth.authToken.file = "/file";
    signingKeyFile = "/file";
    authTokenFile = "/file";
}
```
fails

### 6
```nix
{
    enable = true;
    name = "foo";
    # auth.signingKey.file = "/file";
    # auth.authToken.file = "/file";
    # signingKeyFile = "/file";
    authTokenFile = "/file";
}
```
succeeds

### 7
```nix
{
    enable = true;
    name = "foo";
    # auth.signingKey.file = "/file";
    # auth.authToken.file = "/file";
    signingKeyFile = "/file";
    # authTokenFile = "/file";
}
```
succeeds

### 8
```nix
{
    enable = true;
    name = "foo";
    # auth.signingKey.file = "/file";
    # auth.authToken.file = "/file";
    # signingKeyFile = "/file";
    # authTokenFile = "/file";
}
```
fails, slightly differently, but fails

these show that we only allow the valid combinations. Furthermore I've redirected the `authTokenFile` and `signingKeyFile` into `auth.{authToken,signingKey}.file` which is an `attrTag` which correctly captures the semantics of only one being set being valid. The redirect happens through custom code, as `mkRenamedOptionModule` doesn't play well with `attrTag`. I've made sure that the user is warned about the use of `authTokenFile` and `signingKeyFile` properly.

The only thing I haven't tested is actually running builbot with these changes, since I do not have access to Cachix, but I've confirmed that evaluation behaves as we want. Therefore if someone could test this and run it with Cachix before we merge and have to cut *another* release, it would be greatly appreciated :)

### DIFF
--- a/nix/master.nix
+++ b/nix/master.nix
@@ -8,8 +8,6 @@ let
   cfg = config.services.buildbot-nix.master;
   inherit (lib) mkRemovedOptionModule mkRenamedOptionModule;
 
-  optionsCachix = options.services.buildbot-nix.master.cachix;
-
   interpolateType =
     lib.mkOptionType {
       name = "interpolate";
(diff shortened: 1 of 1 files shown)


MagicRB:
oh and i need to fix the checks.

Mic92:
@mergify queue

PR #243, opened 2024-07-28
feat!: Update output path writing

Minion3665:
- Allow pull requests to create output paths

Previously, it was not possible to get unmerged pull request outputs. I would like to access these (e.g. for hosting a preview version of a website in a similar way to https://docs.netlify.com/site-deploys/deploy-previews/ or https://vercel.com/docs/deployments/preview-deployments)

These are now surfaced under `{owner}/{repo}/pulls/{pr number}`

- Fix repository attribute name conflicts

Previously there was no difference between samely-named attributes in different repositories.

This has been changed so outputs are under `{owner}/{repo}/{branch}`

- Allow full attribute names, still stripping path traversal

Previously, presumably to prevent path traversal, if your attribute name contained slashes buildbot-nix would only take the last segment as an output.

This has been replaced by interpreting slashes as subdirectories and refusing any segments which don't descend into a subdirectory (i.e. are attempting path traversal)

- Still create outputs on skipped builds

Previously when something was skipped, for example a build that was completed in a pull request, the output wouldn't be updated. This made the outputs directory quite unreliable.

Outputs will now always be updated, no matter whether a build was actually executed.

BREAKING-CHANGE: This stops old output locations being outputted. If you rely on these locations, you will need to update whatever relies on them.

### DIFF
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -310,16 +310,54 @@ class UpdateBuildOutput(steps.BuildStep):
         self.project = project
         self.path = path
 
+    def join_traversalsafe(self, root: Path, joined: Path) -> Path:
+        root = root.resolve()
+
+        for part in joined.parts:
+            new_root = (root / part).resolve()
+
+            if not new_root.is_relative_to(root):
+                msg = f"Joined path attempted to traverse upwards when processing {root} against {part} (gave {new_root})"
+                raise ValueError(msg)
+
+            root = new_root
+
+        return root
+
+    def join_all_traversalsafe(self, root: Path, *paths: Path) -> Path:
+        for path in paths:
+            root = self.join_traversalsafe(root, path)
+
+        return root
+
+    @defer.inlineCallbacks
     def run(self) -> Generator[Any, object, Any]:
         props = self.build.getProperties()
-        if props.getProperty("branch") != self.project.default_branch:
+
+        pr = props.getProperty("pr_number")
+
+        if not pr and props.getProperty("branch") != self.project.default_branch:
             return util.SKIPPED
 
-        attr = Path(props.getProperty("attr")).name
+        owner = Path(props.getProperty("owner"))
+        repo = Path(props.getProperty("repository_name"))
+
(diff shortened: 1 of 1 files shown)


zimbatm:
:warning: If arbitrary HTML is rendered in this context, arbitrary JavaScript can be loaded on the same origin as the buildbot instance. This opens up the instance to CSRF and other forms of attack.

Minion3665:
> :warning: If arbitrary HTML is rendered in this context, arbitrary JavaScript can be loaded on the same origin as the buildbot instance. This opens up the instance to CSRF and other forms of attack.

(!!) I must be misunderstanding something here... unless you're talking about my deploy preview idea (which I will implement separately, but thanks for the warning :)) I'm not sure how this PR introduces arbitrary HTML rendering

zimbatm:
You're right. I misunderstood what this PR was about. Looking at the implementation, it changes some routes, and I don't see any obvious security issue here.

MagicRB:
Looks good to me too, @Mic92 do we merge?

Minion3665:
I've added another commit fixing the issue, thanks very much for your patience!

Mic92:
testing.

buildbot_nix/__init__.py
```diff
@@ -310,16 +310,58 @@ def __init__(self, project: GitProject, path: Path, **kwargs: Any) -> None:
...
         out_path = props.getProperty("out_path")
-        # XXX don't hardcode this
-        self.path.mkdir(parents=True, exist_ok=True)
-        (self.path / attr).write_text(out_path)
+
+        if not out_path:  # if, e.g., the build fails and doesn't produce an output
+            return util.SKIPPED
+
+        owner = Path(props.getProperty("owner"))
+        repo = Path(props.getProperty("repository_name"))
```
Mic92, line 348: Those properties only exist in gitea now. Projectname does however: https://github.com/nix-community/buildbot-nix/pull/272/files

However there is also no `pr_number` with the github backend, which has pull requests are skipped just now.

Mic92:
We applied parts of this pull request in https://github.com/nix-community/buildbot-nix/pull/288
If you still want this to be done for pull request, please make a new pull request and also take the properties into account that exist for github. Thanks!

PR #244, opened 2024-07-28
chore(treefmt): Replace root file

Minion3665:
.git/config won't be present when using a VCS such as [jujutsu](https://github.com/martinvonz/jj), however there will always be exactly 1 flake.nix at the root of the repository so we can use it instead

(an example of this being used is also in the treefmt README in the flakes section: https://github.com/numtide/treefmt-nix?tab=readme-ov-file#flakes)

### DIFF
--- a/nix/treefmt/flake-module.nix
+++ b/nix/treefmt/flake-module.nix
@@ -4,7 +4,7 @@
   ];
   perSystem = { pkgs, ... }: {
     treefmt = {
-      projectRootFile = ".git/config";
+      projectRootFile = "flake.nix";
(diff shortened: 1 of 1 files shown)


Minion3665:
> ⚠️ The sha of the head commit of this PR conflicts with #243. Mergify cannot evaluate rules on this PR. ⚠️

That's an exciting error! I'll rebase this once that's in, I guess :)

Mic92:
I am using the license file because we might get flake.nix files when we add integration tests: https://github.com/nix-community/buildbot-nix/pull/247

PR #247, opened 2024-07-28
chore(treefmt): Replace root file

Mic92:
.git/config won't be present when using a VCS such as [jujutsu](https://github.com/martinvonz/jj), however there will always be exactly 1 flake.nix 2 at the root of the repository so we can use it instead

(an example of this being used is also in the treefmt README in the flakes section: https://github.com/numtide/treefmt-nix?tab=readme-ov-file#flakes)

### DIFF
--- a/examples/attic-watch-store.nix
+++ b/examples/attic-watch-store.nix
@@ -1,11 +1,13 @@
-{ pkgs
-, config
-, ...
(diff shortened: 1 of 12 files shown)


Mic92:
@mergify queue

PR #249, opened 2024-07-31
Use installation tokens for GitHub change hooks

MagicRB:
This should fix handling PRs, as reported by @Mic92:
```
Failed fetching PR commit message: response code 401: b'{"message":"Bad credentials","documentation_url":"[https://docs.github.com/rest","status":"401](https://docs.github.com/rest%22,%22status%22:%22401)"}'
```

### DIFF
--- a/buildbot_nix/github_projects.py
+++ b/buildbot_nix/github_projects.py
@@ -291,7 +291,7 @@ class GithubAuthBackend(ABC):
         pass
 
     @abstractmethod
-    def get_repo_token(self, repo: RepoData) -> RepoToken:
+    def get_repo_token(self, repo_full_name: str) -> RepoToken:
         pass
 
     @abstractmethod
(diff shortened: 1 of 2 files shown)


MagicRB:
Changed to use the upstreamed patch, not sure when that will make it into a release, but when it does we'll know as the patch we have won't apply :D

buildbot_nix/github_projects.py
```diff
@@ -585,10 +585,15 @@ def create_reporter(self) -> ReporterBase:
...
     def create_change_hook(self) -> dict[str, Any]:
+        def get_github_token(props: Properties) -> str:
+            return self.auth_backend.get_repo_token(
+                props.getProperty("full_name")
+            ).get()
+
         return {
             "secret": self.webhook_secret,
             "strict": True,
-            "token": self.auth_backend.get_general_token().get(),
```
Mic92, line 591: Do we still need this method afterwards?

MagicRB: Ah good catch we don't, is there some "dead code" thing for python? Would require us to keep explicit exports, which is doable

PR #251, opened 2024-08-01
Support fully private mode

MagicRB:
This places `oauth2-proxy` in front of buildbot and protects the whole instance unconditionally. Buildbot doesn't really support hiding information currently, the websocket is completely unauthenticated and leaks information left and right, therefore a proxy is necessary.

# TODO
- [x] use HTTP basic auth to auth the oauth2 proxy against buildbot and prevent other programs which may have access to `127.0.0.1` from bypassing the oauth proxy.

### DIFF
--- a/nix/master.nix
+++ b/nix/master.nix
@@ -25,6 +25,12 @@ let
       "util.Interpolate(${builtins.toJSON value.value})"
     else
       builtins.toJSON value;
+
+  backendPort =
+    if (cfg.accessMode ? "fullyPrivate") then
+      cfg.accessMode.fullyPrivate.port
+    else
+      config.services.buildbot-master.port;
 in
 {
   imports = [
@@ -226,6 +232,14 @@ in
                     A list of users that should be given access to BuildBot.
                   '';
                   default = [];
+                };
+
+                port = lib.mkOption {
+                  type = lib.types.port;
+                  description = ''
+                    Port number at which the `oauth2-proxy' will listen on.
+                  '';
+                  default = 8020;
                 };
               };
             };
@@ -680,11 +694,6 @@ in
             (cfg.buildbotNixpkgs.python3.pkgs.callPackage ../default.nix { })
             buildbot-gitea
           ];
-
-
-        extraConfig = ''
-          c['www'] = { "port": "unix:/run/buildbot-master/www.unix" }
-        '';
       };
 
       systemd.services.buildbot-master = {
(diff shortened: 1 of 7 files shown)


nix/master.nix
```diff
@@ -680,6 +765,61 @@ in
         ++ lib.optional (cfg.outputsPath != null)
           # Allow buildbot-master to write to this directory
           "d ${cfg.outputsPath} 0755 buildbot buildbot - -";
+
+
+      services.oauth2-proxy = lib.mkIf (cfg.accessMode ? "fullyPrivate") {
```
Mic92, line 770: When this is fleshed out, it would be also nice to have something mentioned in the README or examples?

MagicRB: Of course, I always leave that as last, since rewriting it 3 times is not fun.

MagicRB:
@Mic92 you think this approach is good? I managed to also do authentication with buildbot so that itll autologin based on whatever oauth2-proxy says. It doesn't hide all information if something can bypass the proxy but it protects from something gaining administrator on buildbot just by bypassing the proxy.

Mic92:
> @Mic92 you think this approach is good? I managed to also do authentication with buildbot so that itll autologin based on whatever oauth2-proxy says. It doesn't hide all information if something can bypass the proxy but it protects from something gaining administrator on buildbot just by bypassing the proxy.

I think we should document this as a known issue, but if a system only has private repositories, this is an acceptable tradeof.

MagicRB:
Yeah sounds reasonable, ill write the docs and tests once i finish the implementation

MagicRB:
@zimbatm can you check whether this fulfills the requirements you were looking for? I made sure to protect buildbot with the http basic auth between buildbot and `oauth2-proxy` to prevent rogue FODs from bypassing the authentication and leaking information.

zimbatm:
Yes, that's a good compromise. Do you know if it lets the WebHooks trough unauthenticated?

MagicRB:
Oh yeah it can, i ommited that, oops, ill do it this evening. (It's just setting a url prefix which can go through unauthenticated)

MagicRB:
done. I also added that it will throw back a 401 on `^/api` instead of a redirect.

MagicRB:
https://buildbot.redalder.org/api gives back a 401, not a redirect
https://buildbot.redalder.org/change_hook/github gives back a 400, bypassing authentication
https://buildbot.redalder.org/ redirects to the login page

I added @zimbatm and @Mic92 to my config, so it should let you in as ordinary users if you want to see how it behaves.

zimbatm (approved):

MagicRB:
I've verified this to work with Gitea too, @zimbatm @Mic92 can one of you merge? I've also checked the previous authentication flow, which still works. so no regressions hopefully, yay!

PR #255, opened 2024-08-12
Cache failed builds

MagicRB:
This PR begins pulling in the changes Lix made in regards to the build scheduler. The hope is to make this work, refactor and then do the failed build caching (if this code doesn't do it already, I honestly cannot tell). 

The code from Lix will undergo a lot of changes as it is very untyped and in my humble opinion somewhat hard to read and comprehend.  The Lix folk are of course free to pull any changes we make back into the Lix fork. Though long term my hope would be that we can once again merge the two forks...

### DIFF
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -9,6 +9,8 @@ from collections import defaultdict
 from collections.abc import Generator
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, cast, Tuple
+from dataclasses import dataclass
+import dataclasses
 
 from buildbot.config.builder import BuilderConfig
 from buildbot.configurators import ConfiguratorBase
@@ -61,6 +63,7 @@ from .models import (
     CacheStatus,
     NixDerivation,
 )
+from . import failed_builds
 from .oauth2_proxy_auth import OAuth2ProxyAuth
 from .projects import GitBackend, GitProject
 
@@ -85,10 +88,32 @@ class BuildTrigger(steps.BuildStep):
     builds_scheduler: str
     skipped_builds_scheduler: str
     failed_eval_scheduler: str
+    cached_failure_scheduler: str
     dependency_failed_scheduler: str
     result_list: list[int]
     ended: bool
     waitForFinishDeferred: defer.Deferred[tuple[list[int], int]] | None
+
+    @dataclass
+    class ScheduledJob:
+        job: NixEvalJobSuccess
+        builder_ids: dict[int, int]
+        results: defer.Deferred[list[int]]
+
+        def __iter__(self) -> Generator[Any, None, None]:
+            for field in dataclasses.fields(self):
+                yield getattr(self, field.name)
+
+    @dataclass
+    class DoneJob:
+        job: NixEvalJobSuccess
+        builder_ids: dict[int, int]
+        results: list[int]
+
+        def __iter__(self) -> Generator[Any, None, None]:
+            for field in dataclasses.fields(self):
+                yield getattr(self, field.name)
+
 
     def __init__(
         self,
@@ -97,6 +122,7 @@ class BuildTrigger(steps.BuildStep):
         skipped_builds_scheduler: str,
         failed_eval_scheduler: str,
         dependency_failed_scheduler: str,
+        cached_failure_scheduler: str,
         successful_jobs: list[NixEvalJobSuccess],
         failed_jobs: list[NixEvalJobError],
         report_status: bool,
@@ -112,6 +138,7 @@ class BuildTrigger(steps.BuildStep):
         self.builds_scheduler = builds_scheduler
         self.skipped_builds_scheduler = skipped_builds_scheduler
         self.failed_eval_scheduler = failed_eval_scheduler
+        self.cached_failure_scheduler = cached_failure_scheduler
         self.dependency_failed_scheduler = dependency_failed_scheduler
         self._result_list: list[int] = []
         self.ended = False
@@ -147,13 +174,28 @@ class BuildTrigger(steps.BuildStep):
         source = f"nix-eval-nix"
 
         props = Properties()
-        props.setProperty("virtual_builder_name", job.attr, source)
+        props.setProperty("virtual_builder_name", f".#checks.{job.attr}", source)
         props.setProperty("status_name", f"nix-build .#checks.{job.attr}", source)
         props.setProperty("virtual_builder_tags", "", source)
         props.setProperty("error", job.error, source)
         props.setProperty("attr", job.attr, source)
 
         return (self.failed_eval_scheduler, props)
+
+    def schedule_cached_failure(
+        self,
+        build_props: Properties,
+        job: NixEvalJobSuccess
+    ) -> Tuple[BaseScheduler, Properties]:
+        source = f"nix-eval-nix"
+
+        props = Properties()
+        props.setProperty("virtual_builder_name", f".#checks.{job.attr}", source)
+        props.setProperty("status_name", f"nix-build .#checks.{job.attr}", source)
+        props.setProperty("virtual_builder_tags", "", source)
+        props.setProperty("attr", job.attr, source)
+
+        return (self.cached_failure_scheduler, props)
 
     def schedule_dependency_failed(
         self,
@@ -164,7 +206,7 @@ class BuildTrigger(steps.BuildStep):
         source = f"nix-eval-nix"
 
         props = Properties()
-        props.setProperty("virtual_builder_name", job.attr, source)
+        props.setProperty("virtual_builder_name", f".#checks.{job.attr}", source)
         props.setProperty("status_name", f"nix-build .#checks.{job.attr}", source)
         props.setProperty("virtual_builder_tags", "", source)
         props.setProperty("dependency.attr", dependency.attr, source)
@@ -176,7 +218,7 @@ class BuildTrigger(steps.BuildStep):
         source = f"nix-eval-nix"
 
         props = Properties()
-        props.setProperty("virtual_builder_name", f"hydraJobs.{job.attr}", source)
+        props.setProperty("virtual_builder_name", f".#checks.{job.attr}", source)
         props.setProperty("status_name", f"nix-build .#checks.{job.attr}", source)
         props.setProperty("virtual_builder_tags", "", source)
 
@@ -290,7 +332,7 @@ class BuildTrigger(steps.BuildStep):
             old_paths = list(failed_paths)
             for build in list(jobs):
                 deps: set[str] = job_closures.get(build.drvPath, set())
-                print("deps: " + str(deps))
+
                 for path in old_paths:
                     if path in deps:
                         failed_checks.append(build)
@@ -310,6 +352,7 @@ class BuildTrigger(steps.BuildStep):
         ss_for_trigger = self.prepareSourcestampListForTrigger()
         scheduler_log: Log = yield self.addLog("scheduler")
 
+        # inject failed buildsteps for any failed eval jobs we got
         overall_result = SUCCESS if not self.failed_jobs else util.FAILURE
         for failed_job in self.failed_jobs:
             scheduler_log.addStdout(f'{failed_job.attr} failed eval');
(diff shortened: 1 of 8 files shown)


MagicRB:
Update, it works :tada: it needs a lot of cleanup, typing and general refactoring. And I probably broke combined builds, but hey, it works.

MagicRB:
Todays progress is good, it type checks now, omits errored evals and skips the already cached ones. Further refactoring will allow to show failed build steps for failed evals. And then a failed build cache will achieve failed build caching.

buildbot_nix/__init__.py
```diff
@@ -59,85 +83,218 @@ def __init__(
...
+    def getSchedulerByName(self, name: str) -> BaseScheduler:
+        schedulers = self.master.scheduler_manager.namedServices
+        if name not in schedulers:
+            raise ValueError(f"unknown triggered scheduler: {repr(name)}")
+        sch = schedulers[name]
+        # todo: check ITriggerableScheduler
+        return sch
+
+    def schedule_one(self, build_props: Properties, job: NixEvalJobSuccess) -> Tuple[BaseScheduler, Properties]:
+        source = f"nix-eval-lix"
```
Mic92, line 129:
```suggestion
        source = f"nix-eval-nix"
```

buildbot_nix/__init__.py
```diff
@@ -59,85 +83,218 @@ def __init__(
...
+                    build_schedule_order.append(builds_to_schedule[i])
+                    del builds_to_schedule[i]
+                else:
+                    i += 1
+
+        done = []
+        scheduled: list[Tuple[NixEvalJobSuccess, dict[int, int], defer.Deferred[list[int]]]] = []
+        all_results = SUCCESS
+        ss_for_trigger = self.prepareSourcestampListForTrigger()
+        while len(build_schedule_order) > 0 or len(scheduled) > 0:
```
Mic92, line 220: Is there maybe a way to split this function up into smaller pieces?

Mic92:
@MagicRB Do we still need the combined build if the success status for the `buildbot/nix-eval` could be used as well?

buildbot_nix/__init__.py
MagicRB: Indeed, i havent done that yet, i want to get the features working then refactor more. Introduce more datatypes and split it up

MagicRB:
Thats a separate PR after this one imo, but yea that could be done

MagicRB:
I found a bug in the original lix code, seems that job failures weren't properly propagated to dependents, 
```
    failed_checks: list[NixEvalJob] = []
    failed_paths: list[str] = []
```
need to be initialized to the job that failed or the function will never "recurse" (it's an iterative graph walk which is expressed way more naturally with recursion, but python is pythonic). It appears that I missed this [commit](https://git.lix.systems/lix-project/buildbot-nix/commit/9a15348984a63e965a5b931ce7e51d9f0751c9d6) when doing the initial cherry pick

buildbot_nix/__init__.py
```diff
@@ -49,95 +73,253 @@ class BuildbotNixError(Exception):
...
+    def getFailedDependents(job: NixEvalJobSuccess, jobs: list[NixEvalJobSuccess], job_closures: dict[str, set[str]]) -> list[NixEvalJobSuccess]:
+        jobs = list(jobs)
+        failed_checks: list[NixEvalJob] = [job]
+        failed_paths: list[str] = [job.drvPath]
+        removed = []
+        while True:
+            old_paths = list(failed_paths)
+            for build in list(jobs):
+                deps: set[str] = job_closures.get(build.drvPath, set())
+                print("deps: " + str(deps))
```
Mic92, line 230: replace with log function

buildbot_nix/__init__.py
```diff
@@ -49,95 +73,253 @@ class BuildbotNixError(Exception):
...
+            k: self.getDerivationClosure(k, derivations_inputs).intersection(job_set) for k in job_set
+        }
+        build_schedule_order = self.sortJobsByClosures(self.successful_jobs, job_closures)
+
+        done = []
+        scheduled: list[Tuple[NixEvalJobSuccess, dict[int, int], defer.Deferred[list[int]]]] = []
+        all_results = SUCCESS
+        ss_for_trigger = self.prepareSourcestampListForTrigger()
+        while build_schedule_order or scheduled:
+            print('Scheduling..')
```
Mic92, line 265: same as above.

buildbot_nix/__init__.py
```diff
@@ -49,95 +73,253 @@ class BuildbotNixError(Exception):
...
+            schedule_now = []
+            for build in list(build_schedule_order):
+                if not job_closures.get(build.drvPath, []):
+                    build_schedule_order.remove(build)
+                    schedule_now.append(build)
+            if not schedule_now:
+                print('    No builds to schedule found.')
```
Mic92, line 272: same as above.

buildbot_nix/__init__.py
```diff
@@ -49,95 +73,253 @@ class BuildbotNixError(Exception):
...
+            for job in schedule_now:
+                print(f"    - {job.attr}")
```
Mic92, line 274: same as above.

buildbot_nix/__init__.py
```diff
@@ -49,95 +73,253 @@ class BuildbotNixError(Exception):
...
+                    yield self.addLogWithException(e)
+                    # results = EXCEPTION
+                scheduled.append((job, brids, resultsDeferred))
+
+
+                for brid in brids.values():
+                    url = getURLForBuildrequest(self.master, brid)
+                    yield self.addURL(f"{scheduler.name} #{brid}", url)
+                    self._add_results(brid)
+            print('Waiting..')
```
Mic92, line 299: Same as above.

buildbot_nix/__init__.py
```diff
@@ -49,95 +73,253 @@ class BuildbotNixError(Exception):
...
+            job, brids, _ = scheduled[index]
+            done.append((job, brids, results))
+            del scheduled[index]
+            result = results[0]
+            print(f'    Found finished build {job.attr}, result {util.Results[result].upper()}')
+            if result != SUCCESS:
+                removed = self.getFailedDependents(job, build_schedule_order, job_closures)
+                for job in removed:
+                    build_schedule_order.remove(job)
+                print('    Removed jobs: ' + ', '.join([ job.drvPath for job in removed ]))
```
Mic92, line 313: Same as above.

buildbot_nix/__init__.py
```diff
@@ -49,95 +73,253 @@ class BuildbotNixError(Exception):
...
+            all_results = worst_status(result, all_results)
+            print(f'    New result: {util.Results[all_results].upper()}')
```
Mic92, line 315: Same as above.

buildbot_nix/__init__.py
```diff
@@ -49,95 +73,253 @@ class BuildbotNixError(Exception):
...
+            for dep in job_closures:
+                if job.drvPath in job_closures[dep]:
+                    job_closures[dep].remove(job.drvPath)
+        print('Done!')
```
Mic92, line 319: Same as above.

buildbot_nix/__init__.py
MagicRB: I'm considering replacing with a separate log window in the webui so that the information is easily accessible to users not just the sysadmin.

Mic92: Also works for me.

MagicRB: Im planning to grep the codebase for both lix and print 🤣

MagicRB:
image

Now it properly reports failed dependents and failed evaluations.

and the log has been moved into buildbot's UI

image

Mic92:
Nice.

buildbot_nix/__init__.py
```diff
@@ -14,29 +16,43 @@
...
 
+from pydantic import TypeAdapter
 from twisted.internet import defer
 from twisted.logger import Logger
+from twisted.python.failure import Failure
 
-from .common import (
-    slugify_project_name,
-)
+from . import failed_builds
```
Mic92, line 41: Did you forgot to commit this?

buildbot_nix/__init__.py
```diff
@@ -49,95 +65,439 @@ class BuildbotNixError(Exception):
...
+        while build_schedule_order or scheduled:
+            scheduler_log.addStdout("Scheduling...\n")
+
+            # check which jobs should be scheduled now
+            schedule_now = []
+            for build in list(build_schedule_order):
+                if job_closures.get(build.drvPath):
+                    pass
+                elif (
+                    failed_builds.check_build(build.drvPath)
```
Mic92, line 398: Could we also store timestamp in there for the future? Than we can garbage collect at some point.

Mic92: Maybe use something like json as the value i.e. `{"ts": 1233e131313131}` than we can also add fields.

MagicRB: Could be done yes, will ad before merging

buildbot_nix/__init__.py
MagicRB: Yeah probably

buildbot_nix/failed_builds.py
```diff
@@ -0,0 +1,61 @@
...
+    pass
+
+
+class FailedBuild(BaseModel):
+    derivation: str
+    time: datetime
+
+
+DB_NOT_INIT_MSG = "Database not initialized"
+
```
Mic92, line 24: We can avoid the error above above by adding a database class:

``` 
class Database:
  def __init__(db_path: Path): -> None:
    self.handle = dbm.open(str(db_path), "c")

  def add_build(self, derivation: str, time: datetime) -> None:
    self.handle[derivation] = FailedBuild(derivation=derivation, time=time).model_dump_json()
  # ...
```

Mic92: Looks overall like it would be less code, we just need to pass down the database into the the eval steps where we need it.

MagicRB: I wanted to avoid threading it through for now as we plan to put the data into postgres anyway. And then we'd be accessing it through buildbots machinery hopefully. Treating this as a temporary solution so tried to make it as self contained as possible

MagicRB:
I think this is ready for review, it works I've tested it thoroughly and it seems to behave well.

MagicRB:
@puckipedia the code is different from how you had it, very different, but still in the spirit of cooperation if you want the changes feel free to pull them in. My changes don't actually change any behavior IIRC they just make the code imho better.

If you do want to pull it in, even partially, and something is unclear feel free to contact me on Matrix and I'll help out.

buildbot_nix/failed_builds.py
Mic92: I don't see what you mean by threading?

MagicRB: I meant adding it everywhere only to then remove it later when once we integrate with buildbots database we'll access it through `self.master.db` so we'd be needlessly adding arguments and then removing them. That was my initial thinking, but tbh we've spent more time talking about it than I would have spent doing it 😂

zowoq:
Couple of build retries remnants:

```diff
diff --git a/buildbot_nix/models.py b/buildbot_nix/models.py
index eea11d8..9de66ad 100644
--- a/buildbot_nix/models.py
+++ b/buildbot_nix/models.py
@@ -174,7 +174,6 @@ class PostBuildStep(BaseModel):
 class BuildbotNixConfig(BaseModel):
     db_url: str
     auth_backend: AuthBackendConfig
-    build_retries: int
     cachix: CachixConfig | None
     gitea: GiteaConfig | None
     github: GitHubConfig | None
diff --git a/nix/master.nix b/nix/master.nix
index 3bbd0d6..f72bdf7 100644
--- a/nix/master.nix
+++ b/nix/master.nix
@@ -110,12 +110,6 @@ in
         '';
       };
 
-      buildRetries = lib.mkOption {
-        type = lib.types.int;
-        default = 1;
-        description = "Number of times a build is retried";
-      };
-
       postBuildSteps = lib.mkOption {
         default = [ ];
         description = ''
@@ -606,7 +600,6 @@ in
                 (pkgs.formats.json { }).generate "buildbot-nix-config.json" {
                   db_url = cfg.dbUrl;
                   auth_backend = cfg.authBackend;
-                  build_retries = cfg.buildRetries;
                   cachix =
                     if !cfg.cachix.enable then
                       null
```

MagicRB:
@zowoq good catch, ill rip those out too. Might be a good idea to enable strict deserialization for pydantic so it wouldnt ignore unknown fields.

Mic92:
@mergify rebase

buildbot_nix/__init__.py
```diff
@@ -50,95 +64,350 @@ class BuildbotNixError(Exception):
...
+        return r.difference([derivation])
+
+    @staticmethod
+    def sort_jobs_by_closures(
+        jobs: list[NixEvalJobSuccess], job_closures: dict[str, set[str]]
+    ) -> list[NixEvalJobSuccess]:
+        sorted_jobs = []
+        sorter = graphlib.TopologicalSorter(job_closures)
+
+        for item in sorter.static_order():
```
MagicRB, line 279: Sort the jobs directly not the strings

MagicRB:
@Mic92 I cherry picked your changes and forced pushed what I had locally, (which was a lot, apologies about that).

MagicRB:
Fully functional, tested on both Forgejo and GitHub, utilizes the new code in `nix-eval-jobs` which speeds up the build significantly.

Mic92:
I test again.

Mic92:
> @Mic92 git is failing me here, what changes are you making? :)

It's a rebase + https://github.com/nix-community/buildbot-nix/pull/255/commits/54961b30696670e7b973c4270b2b46c85b8ec987

zowoq:
current

Just noticed that buildbot UI is missing the project details that was there previously:

original

MagicRB:
Ah good catch, probably just the props being wrong, ill fix today

Mic92:
I think this also have been wrong on gitea as it is also show github.

MagicRB:
```python
   if repo_name is not None:
        name = f"github:{repo_name}#checks.{name}"
    else:
        name = f"checks.{name}"
 ```
 ah we're missing this bit of code, which was wrong anyway. I'll devise a fix

MagicRB:
@Mic92 can we open issues for follow ups? Cause it is almost impossible to track with all the flake update PRs. They create way too much noise.

PR #260, opened 2024-08-20
Different combined reports

MagicRB:


MagicRB:
With this new version, it now uses custom buildbot message queue event types, which completely bypasses any issues we might have with buildbot not supporting the specific reports we might want.

MagicRB:
Dependent on #255, tested to be fully functional both on Forgejo and GitHub, combined and uncombined bulids.

zowoq:
Just want to confirm as I'll need to announce it for the community buildbot: 

This PR was a breaking change, `buildbot/nix-build` is now the meta "everything builds successfully" check instead of `buildbot/nix-eval`?

Mic92:
Yes. We will make this a major release. It will break existing mergify configuration, but in the long-term it gives us saner notifications.

Mic92:
The mergify editor hopefully makes the transition soon as it allows to make pull request.

Mic92:
You can also send out pull request before doing the upgrade probably. It would just disable merge queues for a little bit.

MagicRB:
Do we cut now or wait for effects? I may be able to get them done over the weekend

Mic92:
I'll probably will not get to cut a release until next week. So if it is mergeable by next week, it will get in, otherwise 1.0 will go out without it.

MagicRB:
Roger

zowoq:
Don't seem to get a "everything builds successfully" `buildbot/nix-build` check if jobReportLimits > number of builds?

MagicRB:
@zowoq yeah that was by design, is it needed for better integration? like mergify and stuff. It's easy to add back in (removing 2 ifs).

zowoq:
No, isn't needed, just something I noticed.

MagicRB:
Ok, its very easy to add back, or gate behind a switch, so lmk if you want it back.

Mic92:
I think I am now also for removing jobReportLimits and make only two notifications the default.

MagicRB:
Id be in favor of it being default, but i do want the verbose version for my own things

Mic92:
What I found confusing is that I wouldn't get the generic "build" notification if I increase jobReportLimits to non-zero. This makes it hard to have required checks.

PR #263, opened 2024-08-25
drop delete legacy gcroot

zowoq:


nix/master.nix
```diff
@@ -778,11 +778,7 @@ in
       };
 
       systemd.tmpfiles.rules =
-        [
-          # delete legacy gcroot location, can be dropped after 2024-06-01
-          "R /var/lib/buildbot-worker/gcroot - - - - -"
-        ]
-        ++ lib.optional (cfg.outputsPath != null)
+        lib.optional (cfg.outputsPath != null)
```
zowoq, line 781: @Mic92 is `outputsPath` feature not working? It looks like it is enabled in your dotfiles but https://buildbot.thalheim.io/nix-outputs is empty?

Mic92:
@mergify queue

nix/master.nix
Mic92: Looks like the nginx configuration broke. I have a look. But I also just discovered a different problem. I need to introduce namespacing in there.

zowoq: > I need to introduce namespacing in there.

https://github.com/nix-community/buildbot-nix/pull/243

Looks like this PR handles namespacing, PRs and skipped builds.

PR #272, opened 2024-09-12
feat!: Update output path writing

Mic92:
- Allow pull requests to create output paths

Previously, it was not possible to get unmerged pull request outputs. I would like to access these (e.g. for hosting a preview version of a website in a similar way to https://docs.netlify.com/site-deploys/deploy-previews/ or https://vercel.com/docs/deployments/preview-deployments)

These are now surfaced under `{owner}/{repo}/pulls/{pr number}`

- Fix repository attribute name conflicts

Previously there was no difference between samely-named attributes in different repositories.

This has been changed so outputs are under `{owner}/{repo}/{branch}`

- Allow full attribute names, still stripping path traversal

Previously, presumably to prevent path traversal, if your attribute name contained slashes buildbot-nix would only take the last segment as an output.

This has been replaced by interpreting slashes as subdirectories and refusing any segments which don't descend into a subdirectory (i.e. are attempting path traversal)

- Still create outputs on skipped builds

Previously when something was skipped, for example a build that was completed in a pull request, the output wouldn't be updated. This made the outputs directory quite unreliable.

Outputs will now always be updated, no matter whether a build was actually executed.

BREAKING-CHANGE: This stops old output locations being outputted. If you rely on these locations, you will need to update whatever relies on them.

fix: Skip writing path with failed builds

Previously, if a build failed to produce any output we would try to write "None" to the output file. This doesn't work, causing the job to error. Instead, we should skip writing the output and preserve the original "failed" status.

### DIFF
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -756,16 +756,57 @@ class UpdateBuildOutput(steps.BuildStep):
         self.project = project
         self.path = path
 
+    def join_traversalsafe(self, root: Path, joined: Path) -> Path:
+        root = root.resolve()
+
+        for part in joined.parts:
+            new_root = (root / part).resolve()
+
+            if not new_root.is_relative_to(root):
+                msg = f"Joined path attempted to traverse upwards when processing {root} against {part} (gave {new_root})"
+                raise ValueError(msg)
+
+            root = new_root
+
+        return root
+
+    def join_all_traversalsafe(self, root: Path, *paths: Path) -> Path:
+        for path in paths:
+            root = self.join_traversalsafe(root, path)
+
+        return root
+
+    @defer.inlineCallbacks
     def run(self) -> Generator[Any, object, Any]:
         props = self.build.getProperties()
-        if props.getProperty("branch") != self.project.default_branch:
+
+        pr = props.getProperty("pr_number")
+
+        if not pr and props.getProperty("branch") != self.project.default_branch:
             return util.SKIPPED
 
-        attr = Path(props.getProperty("attr")).name
         out_path = props.getProperty("out_path")
-        # XXX don't hardcode this
-        self.path.mkdir(parents=True, exist_ok=True)
-        (self.path / attr).write_text(out_path)
+
+        if not out_path:  # if, e.g., the build fails and doesn't produce an output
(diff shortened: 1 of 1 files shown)


buildbot_nix/__init__.py
```diff
@@ -756,16 +756,57 @@ def __init__(self, project: GitProject, path: Path, **kwargs: Any) -> None:
...
+            root = self.join_traversalsafe(root, path)
+
+        return root
+
+    @defer.inlineCallbacks
     def run(self) -> Generator[Any, object, Any]:
         props = self.build.getProperties()
-        if props.getProperty("branch") != self.project.default_branch:
+
+        pr = props.getProperty("pr_number")
```
Mic92, line 783: Ah I forgot, why I couldn't merge this. We don't have this in github.

Mic92:
Reverted in f36faa9d15ef41abd5e4f8184a36754fd2c913c5

buildbot_nix/__init__.py
zowoq: Would be useful to have this for outputs on the default branch, can PRs be dropped? e.g.

```diff
diff --git a/buildbot_nix/__init__.py b/buildbot_nix/__init__.py
index 3dd09f3..1d6dec5 100644
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -780,9 +780,7 @@ class UpdateBuildOutput(steps.BuildStep):
     def run(self) -> Generator[Any, object, Any]:
         props = self.build.getProperties()
 
-        pr = props.getProperty("pr_number")
-
-        if not pr and props.getProperty("branch") != self.project.default_branch:
+        if props.getProperty("branch") != self.project.default_branch:
             return util.SKIPPED
 
         out_path = props.getProperty("out_path")
@@ -792,7 +790,7 @@ class UpdateBuildOutput(steps.BuildStep):
 
         project_name = Path(props.getProperty("projectname"))
 
-        target = Path(f"pulls/{pr}") if pr else Path(props.getProperty("branch"))
+        target = Path(props.getProperty("branch"))
 
         attr = Path(props.getProperty("attr"))
 

```

Mic92: yes. Sounds like a good idea.

PR #281, opened 2024-09-21
Fix missing information in build names

### LINKED ISSUE #278: Build names carry less information
After #255 it appears that the name of individual `nix-build`s has changed, see the following for a comparison.
before:
image
after:
image

This is due to code that didn't quite make it through the refactor intact:
```python
if repo_name is not None:
    name = f"github:{repo_name}#checks.{name}"
else:
    name = f"checks.{name}"
```

MagicRB:
Fixes #278, see images:
image
image


zowoq:
Thanks!

buildbot

PR #282, opened 2024-09-21
Always report `nix-build`, even with individual build reports

### LINKED ISSUE #279: Always report `nix-build`, even with individual build reports
As said by @Mic92:

> What I found confusing is that I wouldn't get the generic "build" notification if I increase jobReportLimits to non-zero. This makes it hard to have required checks.

MagicRB:
Always reports the combined `nix-build` build report, even with individual builds, closes #279, see image:
image


Mic92:
@mergify queue

PR #283, opened 2024-09-21
Remember first failed build URL and display it in cached failures

### LINKED ISSUE #280: Cached failures don't link back to the original failing build
Currently a cached failure doesn't link back to the build that originally failed, only a timestamp is displayed. This makes it hard to actually see why a build is failing.

Quoting @Mic92
>if you get a cached failure, people don't come across the idea that they can look in the previous build for logs. And it is sometimes not even the previous build that has the logs for a failure. Sometimes one has to click several times back.

MagicRB:
Closes #280, currently a draft because it doesn't yet output a clickable link, it merely just dumps a link into the textual output of the cached failed build, see image:
image

Next I will attempt to add a clickable link using HTML.


Mic92:
@mergify queue

Issue #285, opened 2024-09-22
`lib.interpolate` function not fully plumbed through in postBuildSteps.*.environment


antifuchs:
Using `inputs.buildbot-nix.lib.interpolate` with the `postBuildSteps.*.environment` option doesn't result in the call to `util.Interpolate` being made in the resulting configuration; instead, the nix config writes `{"value": "%(prop:attr)s", "_type": "interpolate"}` to the configuration file, which then fails the config test:

with the config file portion generated like this:

```json
  "post_build_steps": [
    {
      "command": "/nix/store/6a496z3csn66d4qfrs8mkaxkycx3asvs-deploy-fly/bin/deploy-fly",
      "environment": {
        "BOINKOR_ATTR": {
          "_type": "interpolate",
          "value": "%(prop:attr)s"
        },
        "BOINKOR_BRANCH": {
          "_type": "interpolate",
          "value": "%(prop:branch)s"
        },
        "BOINKOR_FLYSCRIPT": {
          "_type": "interpolate",
          "value": "%(prop:out_path)s/bin/deploy-fly"
        }
      },
      "name": "Deploy to fly.io"
    }
  ],
```

This is the log message that happens on startup:
```
Sep 22 17:42:10 gloria twistd[398728]: 2024-09-22T17:42:10-0400 [-] error while parsing config file:
Sep 22 17:42:10 gloria twistd[398728]:         Traceback (most recent call last):
Sep 22 17:42:10 gloria twistd[398728]:           File "/nix/store/0bkcrjqrdkhwrvj2574jji8hmrq7mcsm-python3-3.12.5-env/lib/python3.12/site-packages/twisted/python/threadpool.py", line 285, in <lambda>
Sep 22 17:42:10 gloria twistd[398728]:             inContext.theWork = lambda: context.call(  # type: ignore[attr-defined]
Sep 22 17:42:10 gloria twistd[398728]:           File "/nix/store/0bkcrjqrdkhwrvj2574jji8hmrq7mcsm-python3-3.12.5-env/lib/python3.12/site-packages/twisted/python/context.py", line 117, in callWithContext
Sep 22 17:42:10 gloria twistd[398728]:             return self.currentContext().callWithContext(ctx, func, *args, **kw)
[...]

PR #286, opened 2024-09-22
Fix being unable to declare custom `PostBuildStep`s from Nix

MagicRB:
See #285 for context. I tested both Cachix and a custom step to work properly, see [buildbot my instance](https://buildbot.redalder.org/#/builders/56/builds/91).


Mic92:
@mergify queue

PR #288, opened 2024-09-23
Revisited: feat!: Update output path writing

Mic92:
https://github.com/nix-community/buildbot-nix/pull/243

### DIFF
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -784,16 +784,57 @@ class UpdateBuildOutput(steps.BuildStep):
         self.project = project
         self.path = path
 
+    def join_traversalsafe(self, root: Path, joined: Path) -> Path:
+        root = root.resolve()
+
+        for part in joined.parts:
+            new_root = (root / part).resolve()
+
+            if not new_root.is_relative_to(root):
+                msg = f"Joined path attempted to traverse upwards when processing {root} against {part} (gave {new_root})"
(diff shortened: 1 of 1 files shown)


buildbot_nix/__init__.py
```diff
@@ -784,16 +784,55 @@ def __init__(self, project: GitProject, path: Path, **kwargs: Any) -> None:
...
-        (self.path / attr).write_text(out_path)
+
+        if not out_path:  # if, e.g., the build fails and doesn't produce an output
+            return util.SKIPPED
+
+        project_name = Path(props.getProperty("projectname"))
+
+        target = Path(props.getProperty("branch"))
+
+        attr = Path(props.getProperty("attr"))
```
zowoq, line 823: Should this stay the same as it was a few lines above?
```
attr = Path(props.getProperty("attr")).name
``

Comment (Mic92) [reply buildbot_nix/__init__.py:823]: I have a look.

Comment (Mic92) [reply buildbot_nix/__init__.py:823]: I thought about it a bit, and normalizing attributes, could lead to collisions, no?

Comment (Mic92) [reply buildbot_nix/__init__.py:823]: i.e. if someone does something like this:

```
{
  "foo" = ...;
  "foo/bar" = ...;
}

Mic92: Mhm. Also thinkinng about it this specific case would create collisions anyway.

PR #290, opened 2024-09-23
url quote all paths when updating build output paths

Mic92:
Not yet tested. cc @zowoq

### DIFF
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -6,6 +6,7 @@ import json
 import multiprocessing
 import os
 import re
+import urllib.parse
 from collections import defaultdict
 from collections.abc import Generator, Iterable
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


buildbot_nix/__init__.py
```diff
@@ -816,14 +817,13 @@ def run(self) -> Generator[Any, object, Any]:
...
 
-        project_name = Path(props.getProperty("projectname"))
-
-        target = Path(props.getProperty("branch"))
-
-        attr = Path(props.getProperty("attr"))
+        owner = urllib.parse.quote_plus(self.project.owner)
+        repo = urllib.parse.quote_plus(self.project.name)
+        target = urllib.parse.quote_plus(props.getProperty("branch"))
+        attr = urllib.parse.quote_plus(props.getProperty("attr"))
```
Mic92, line 823: This on its own, should already be filesystem safe... The test below is just a safety check.

Mic92:
Seems to work as intended.

PR #291, opened 2024-09-23
[total untested] async syntax

Mic92:
according to twisted documentation, this should work?

### DIFF
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -254,13 +254,12 @@ class BuildTrigger(buildstep.ShellMixin, steps.BuildStep):
             return (self.builds_scheduler, props)
         return (self.skipped_builds_scheduler, props)
 
-    @defer.inlineCallbacks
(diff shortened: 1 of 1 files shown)


MagicRB:
If this worked, it would be sooo good

MagicRB:
@Mic92 you missed some, 
https://github.com/nix-community/buildbot-nix/blob/main/buildbot_nix/__init__.py#L289
https://github.com/nix-community/buildbot-nix/blob/main/buildbot_nix/nix_status_generator.py#L61

zowoq:
Looks like there is something wrong with this, seems the scheduler is only allowing one build at a time?

Mic92:
Indeed. I will run a git bisect.

Mic92:
@zowoq I think I found the issue: https://github.com/nix-community/buildbot-nix/pull/317
I think the old code was incorrect and async/await checks just uncovered it.

Issue #115, opened 2024-04-06
Hercules-effects support


Mic92:

zowoq:
@Mic92 How you you plan to handle secrets? Separate secrets file for each repo or just one secret file per instance like hercules?

Mic92:
Ideally one per repository if I can make it secure. Maybe each repository gets a public key for encryption.

zowoq:
I forgot to ask about the `GitToken` secret, it is a short lived token from the github app set by the hercules backend.

Possible that some repos may not even need to set secrets if they had this token as it could be used for pushing to the repo, pages, releases, etc.

zowoq:
> https://github.com/hercules-ci/hercules-ci-agent/blob/master/docs/modules/ROOT/pages/effects.adoc
> The store and a nix-daemon socket are accessible.

There is some logic for the socket in [ercules-ci-agent/src/Hercules/Effect.hs](https://github.com/hercules-ci/hercules-ci-agent/blob/c0c345b08a5d9f6f37d321141c2876a3a7bc5885/hercules-ci-agent/src/Hercules/Effect.hs), not sure how much of that would need to be replicated.

zowoq:
These are the changes that were needed for a `hci-effects.runNixDarwin` deployment to work. I guess this would be okay while buildbot effects are restricted to running on the default branch?

https://buildbot.nix-community.org/#/builders/3618/builds/24/steps/3/logs/stdio

```diff
diff --git a/buildbot_effects/buildbot_effects/__init__.py b/buildbot_effects/buildbot_effects/__init__.py
index 7b2273e90..c60aa5dcc 100644
--- a/buildbot_effects/buildbot_effects/__init__.py
+++ b/buildbot_effects/buildbot_effects/__init__.py
@@ -190,6 +190,7 @@ def run_effects(
     env["HERCULES_CI_SECRETS_JSON"] = "/run/secrets.json"
     env["NIX_BUILD_TOP"] = "/build"
     env["TMPDIR"] = "/tmp"  # noqa: S108
+    env["NIX_REMOTE"] = "daemon"
     clear_env = set()
     clear_env.add("TMP")
     clear_env.add("TEMP")
@@ -235,6 +236,9 @@ def run_effects(
         "/nix/store",
         "--hostname",
         "hercules-ci",
+        "--bind",
+        "/nix/var/nix/daemon-socket/socket",
+        "/nix/var/nix/daemon-socket/socket",
     ]
 
[...]

PR #307, opened 2024-10-01
Secrets for hercules-ci style effects

antifuchs:
This PR to another PR (sorry for the complexity) is what I use currently to build and deploy my flake-based personal system configuration. This holds all the changes I had to make to get `flyctl deploy` working for docker images built in my flake:

* Add `/var/tmp` as a tmpfs mount point in the bwrapped effects runner
* Flush at the end of writing secrets.json before passing it to `bwrap` (otherwise that file would be empty, thanks python buffering).
* Make per-repo secrets configurable and pass those through from the nix config to the buildbot master plugin, to each worker, and pick that up in the buildbot_effects commandline invocation.

With these changes, I have successfully run a flyctl deploy with the access token sourced from secrets passed in on that per-repo basis.

### DIFF
--- a/buildbot_effects/__init__.py
+++ b/buildbot_effects/__init__.py
@@ -142,12 +142,15 @@ def parse_derivation(path: str) -> dict[str, Any]:
     return json.loads(proc.stdout)
 
 
-def env_args(env: dict[str, str]) -> list[str]:
+def env_args(env: dict[str, str], clear_env: set[str]) -> list[str]:
     result = []
     for k, v in env.items():
         result.append("--setenv")
         result.append(f"{k}")
         result.append(f"{v}")
+    for k in clear_env:
+        result.append("--unsetenv")
+        result.append(f"{k}")
     return result
 
 
@@ -180,6 +183,11 @@ def run_effects(
     env["IN_HERCULES_CI_EFFECT"] = "true"
     env["HERCULES_CI_SECRETS_JSON"] = "/run/secrets.json"
     env["NIX_BUILD_TOP"] = "/build"
+    env["TMPDIR"] = "/tmp"  # noqa: S108
+    clear_env = set()
+    clear_env.add("TMP")
+    clear_env.add("TEMP")
+    clear_env.add("TEMPDIR")
     bwrap = shutil.which("bwrap")
     if bwrap is None:
         msg = "bwrap' executable not found"
@@ -231,7 +239,7 @@ def run_effects(
                 "/run/secrets.json",
             ],
         )
-        bubblewrap_cmd.extend(env_args(env))
+        bubblewrap_cmd.extend(env_args(env, clear_env))
         bubblewrap_cmd.append("--")
         bubblewrap_cmd.extend(sandboxed_cmd)
         with pipe() as (r_file, w_file):
--- a/buildbot_effects/__init__.py
+++ b/buildbot_effects/__init__.py
@@ -223,6 +223,7 @@ def run_effects(
         secrets = secrets.copy()
         secrets["hercules-ci"] = {"data": {"token": "dummy"}}
         tmp.write(json.dumps(secrets).encode())
+        tmp.flush()
         bubblewrap_cmd.extend(
             [
                 "--ro-bind",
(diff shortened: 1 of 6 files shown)


antifuchs:
This PR might address the question around secrets in #115. It's not public-key encrypted (and there remains the question on whether this unduly exposes secrets to forks, if we ever make other branches have effects), but it's a start I think.

buildbot_effects/__init__.py
```diff
@@ -180,6 +180,10 @@ def run_effects(
     env["IN_HERCULES_CI_EFFECT"] = "true"
     env["HERCULES_CI_SECRETS_JSON"] = "/run/secrets.json"
     env["NIX_BUILD_TOP"] = "/build"
+    env["TEMPDIR"] = "/tmp"
+    env["TMPDIR"] = "/tmp"
+    env["TEMP"] = "/tmp"
+    env["TMP"] = "/tmp"
```
Mic92, line 186: I would like to keep only TMPDIR here. I consider the other ones more of a historical artifact.

buildbot_effects/__init__.py
```diff
@@ -203,6 +207,8 @@ def run_effects(
         "--tmpfs",
         "/tmp",  # noqa: S108
         "--tmpfs",
+        "/var/tmp",  # noqa: S108
+        "--tmpfs",
```
Mic92, line 211: What is /var/tmp used for here?

buildbot_effects/__init__.py
antifuchs: Good point. I reset every one that matched `/tmp/nix-shell...` - didn't know which one it was that made this tool behave badly. Made me think, should we re-set the environment and only carry relevant stuff over into the effect, like PATH and the things we set in the runner? There's a _lot_ of crud on the environment from `nix develop`, like stale CFLAGS and so on.

buildbot_effects/__init__.py
antifuchs: Skopeo uses that and doesn't appear to use `$TMPDIR` for that location; I see now that it's possible to override with the `--tmpdir` option, but if one tool gets this wrong, chances are many other tools will.

antifuchs:
Updated to address the TMPDIR comment.

buildbot_effects/__init__.py
Mic92: Do we not have a minimal env here? I don't recall the details. In fact we only should set our own environment variables and inherit nothing implicit.

buildbot_effects/__init__.py
Mic92: Oh, I see:

```
17-// DO NOT change this, instead see unixTempDirForBigFiles above.
18:const builtinUnixTempDirForBigFiles = "/var/tmp"
```

This is very odd behavior. Do they expect people to run this as root?

Mic92: On the other hand, you are not really solving their concern, because you make the directory a tmpfs again. Should we not rather make `/var` be backed by some actual filesystem?

antifuchs: I'm not sure there's a problem to be solved besides making it available - skopeo requires some directory to stage its "big" files, and /var/tmp works just as well as /tmp for those purposes (they're both mode `1777` on non-nixos systems).

If a tool needs a persisted cache, it's probably better to use that "state file" mechanism exposed by hci effects...

Mic92: From what I can read, it actually unpacks docker images, there. So those can be bigger than your RAM. I have the feeling if applications reach out to write to /var/tmp specifically they must have a good reason to not use /tmp.

Mic92: I also want to keep the possibility of using impure derivations at some point instead of bubblewrap because that's more cross-platform. We don't have /var/tmp there as well.

antifuchs: Hm, yeah. I mean, I'm happy to back this out (it's not _necessary_ for my use case), but it feels like `/var/tmp` is a thing that tools might expect of a working system. There should be a way to get one (one way or another) but that can remain a follow-up task after effects land.

antifuchs: Backed out.

PR #308, opened 2024-10-01
Add aarch64-darwin to systems the flake cares about

antifuchs:
This enables the dev shell & corresponding dev tools on recent macs, which happens to be the machine type I'm using (-:


MagicRB:
We first need the checks to pass or be disabled, @antifuchs can you try running `nix flake check` on your end? I am not familiar with darwin *at all*

antifuchs:
Ah, I see; guess the checks need to be configured to run only on the platform we care about as a target, now. I'll see about that.

Mic92:
@mergify rebase

PR #311, opened 2024-10-05
Allows restarting individual builds

### LINKED ISSUE #287: Rebuild a "cached failure" step should try a proper rebuild
While the "rebuild" button works on the whole matrix, it doesn't work, when we select a single build.

MagicRB:
Maybe closes #287

### DIFF
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -13,7 +13,7 @@ from dataclasses import dataclass
 from datetime import UTC, datetime
 from multiprocessing import cpu_count
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Callable
 
 from buildbot.config.builder import BuilderConfig
 from buildbot.configurators import ConfiguratorBase
@@ -982,15 +982,14 @@ async def do_register_gcroot_if(s: steps.BuildStep) -> bool:
     )
 
 
-def nix_build_config(
+def nix_build_steps(
     project: GitProject,
     worker_names: list[str],
     post_build_steps: list[steps.BuildStep],
     outputs_path: Path | None = None,
-) -> BuilderConfig:
-    """Builds one nix flake attribute."""
-    factory = util.BuildFactory()
-    factory.addStep(
+    global_do_step_if: Callable[[steps.BuildStep], bool] | bool = True,
+) -> list[steps.BuildStep]:
+    out_steps = [
         NixBuildCommand(
             env={},
             name="Build flake attr",
@@ -1013,11 +1012,9 @@ def nix_build_config(
             # We increase this over the default since the build output might end up in a different `nix build`.
             timeout=60 * 60 * 3,
             haltOnFailure=True,
+            doStepIf=global_do_step_if,
         ),
-    )
-    factory.addSteps(post_build_steps)
-
-    factory.addStep(
+        *post_build_steps,
         Trigger(
             name="Register gcroot",
             waitForFinish=True,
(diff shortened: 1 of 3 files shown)


MagicRB:
Okay so the status is that one can restart a build without redoing the whole eval->build thing. When such a cached failed build is restart it will mostly behave as if it wasn't cached. Remaining things to do are:
1. fix uncombined reports for these restarts
2. currently cached failed builds run post build steps

buildbot_nix/__init__.py
```diff
@@ -765,11 +713,24 @@ def __init__(self, **kwargs: Any) -> None:
 
     @defer.inlineCallbacks
     def run(self) -> Generator[Any, object, Any]:
+        if self.build.reason == "rebuild" and not self.getProperty("combine_builds"):
+            print("SENDING COMBINED BUILD")
```
Mic92, line 717:
```suggestion
```

buildbot_nix/__init__.py
```diff
@@ -765,11 +713,24 @@ def __init__(self, **kwargs: Any) -> None:
...
+            )
+
         # run `nix build`
         cmd: remotecommand.RemoteCommand = yield self.makeRemoteShellCommand()
         yield self.runCommand(cmd)
 
         res = cmd.results()
+
+        if self.build.reason == "rebuild" and not self.getProperty("combine_builds"):
+            print("SENDING COMBINED BUILD")
```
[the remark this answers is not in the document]
Mic92: I assumed that. But I just added these as a reminder in case they are forgotten.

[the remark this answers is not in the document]
MagicRB: :) i know im just saying

MagicRB:
Update, fixed both points mentioned above, 1. was working just needed verification, 2. fixed in latest commit

buildbot_nix/nix_status_generator.py
```diff
@@ -131,26 +190,37 @@ async def generate(
                 formatter, master, reporter, data
             )
 
-            if event in {"started-nix-eval", "finished-nix-eval"}:
+            if event in {
```
Mic92, line 193: Would it not make sense to convert the string to the our event enum and than return early if the event is not in our enum?

Mic92: I believe enums have a __members__ field or something that can be checked against.

MagicRB: Oh yeah, that makes way more sense, good catch

MagicRB: Done, but untested

Mic92:
@mergify rebase

Mic92:
@MagicRB when applying this change, I am no longer getting any buildbot github updates.

MagicRB:
Which change? The whole PR?

buildbot_nix/nix_status_generator.py
```diff
@@ -131,26 +190,45 @@ async def generate(
...
-                    "generator",
-                )
-            if event in {"finished-nix-eval", "finished-nix-build"}:
-                report["builds"][0]["complete"] = True
-                report["builds"][0]["complete_at"] = datetime.now(tz=UTC)
+            event_typed: CombinedBuildEvent = CombinedBuildEvent(event)
+            match event_typed:
+                case (
+                    CombinedBuildEvent.STARTED_NIX_EVAL
+                    | CombinedBuildEvent.FINISHED_NIX_EVAL
```
Mic92, line 197: Oh, this might be actually the wrong syntax.

MagicRB: Ive no clue how `match` works in python so cant comment

Mic92:
> Which change? The whole PR?

Whole pr. But I just found a bug.

MagicRB:
Pretty sure it worked on my machine :tm:

MagicRB:
"Dont run post build steps on a cached failure" works for sure, im positive i tested that one

PR #312, opened 2024-10-05
Configurable branches

MagicRB:


buildbot_nix/__init__.py
```diff
@@ -276,10 +276,10 @@ async def schedule(
         for brid in brids.values():
             url = getURLForBuildrequest(self.master, brid)
             await self.addURL(f"{scheduler.name} #{brid}", url)
-            self._add_results(brid)
+            await self._add_results(brid)
```
Mic92, line 279: This is now also in main.

MagicRB: Will revert, i just needed pylsp to shur up

nix/master.nix
```diff
@@ -492,6 +492,38 @@ in
...
+          description = ''
+            A `re` Python compatible regex specifying which additional branches should
+            buildbot-nix consider in addition to the default branch. Like for the default
+            branch, buildbot-nix will register GC roots for any builds it performs for this
+            branch. An empty regex means to not consider any additional branches.
+          '';
+          default = "";
+        };
+
+        secondary = lib.mkOption {
```
Mic92, line 517: Do we need to differentiate here? Can we not generate a GC root for all explicitly defined branches?

MagicRB: We could, thats another option

steveej: maybe the easier thing to do for now is to be less opinionated and expose all these knobs in the config. `branches` could be a list of attrs.
because i'm not that well versed on top of my head with the type library, here's a pseudo-code type for the `options.branches`:

`[ { branchRe: str, buildPRs: bool, registerGCRoots: bool, copyOutputs: bool ]`

this gives enough flexibility for all discussed use-cases.

MagicRB: So a list of regex selectors and then flags on each one, dictating what to do

steveej: yes, that's the idea. it could also be an attrset in which the name is a unique identifier instead of anonymous attrs in a list. i've found lists harder to debug.

MagicRB: It'll be in TOML maybe, since im looking into making this per-repo. Though maybe a default list + per-repo would be good

steveej: example

```nix
branches.release = {
   match = "release-.*";
   buildPRs = true;
   registerGCRoots = true;
   copyOutputs = true;
};
```

MagicRB: ``
[branches.release]
match = "release-*"
buildsPRs = true
registerGCRoots = true
copyOutputs = true
```

In a `buildbot-nix.toml` file in the root of the default branch

Comment (steveej) [reply nix/master.nix:517]: > It'll be in TOML maybe, since im looking into making this per-repo. Though maybe a default list + per-repo would be good

i'm a bit wary of attempting per-repo settings in the first iteration because it opens up a lot of security questions. i agree that it's a nice addon for a subsequent iteration.

Comment (steveej) [reply nix/master.nix:517]: > In a `buildbot-nix.toml` file in the root of the default branch

i assume it's possible to transform the nix expression into toml and feed that to the python implementation handling this code that part can be implemented and later be reused for files from the repo.

Comment (MagicRB) [reply nix/master.nix:517]: Well it'll be largely the same code. As in the per-repo `BranchConfig` object gets merged with the global one, *somehow*. So technically the only security critical part is tbe *somehow*. But yeah probanly. I'll finish up #318 without being able to declare branches

File: nix/master.nix [hunk 3]
```diff
@@ -492,6 +492,38 @@ in
         '';
         default = 50;
       };
+
+      branches = {
```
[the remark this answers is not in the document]
MagicRB: #318 is ready and I'm not sure if doing it per-repo but on the instance config side and then switching to per-repo in the repo is worth the trouble. Is there actually a need for instance per-repo or is it just a stopgap solution?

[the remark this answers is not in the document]
zowoq: Just a stopgap, per-repo in the repo is better.

nix/master.nix
MagicRB: I'm proceeding with this design, except that `buildPRs` cannot be implemented the way I took it which is "build PRs based on which branch they're based on, and if that branch matches a glob". Such an implementation is not possible as buildbot doesn't have nor expose this information for a `Change` (see [Change](https://github.com/buildbot/buildbot/blob/master/master/buildbot/changes/changes.py#L33) and [ChangeFilter](https://github.com/buildbot/buildbot/blob/5e9c38f8948f622521420a25f5f250d53f273fe2/master/buildbot/changes/filter.py#L26)). So rather than that I think we just keep `buildPRs` as a instance global option (repository scoped once that's in) and keep the simple behavior with PRs.

MagicRB:
I need a second pair of eyes on [243aa6b97b4b6d062ac85c6c076676ea5acbd3ca](https://github.com/nix-community/buildbot-nix/pull/312/commits/243aa6b97b4b6d062ac85c6c076676ea5acbd3ca) the way I'm reading that is that updating build outputs and gcroots would execute also for PRs, which seems like a recipe for disaster.

buildbot_nix/__init__.py
```diff
@@ -964,24 +974,26 @@ def nix_eval_config(
...
     ).getRenderingFor(s.getProperties())
-    out_path = await util.Property("out_path").getRenderingFor(s.getProperties())
-    default_branch = await util.Property("default_branch").getRenderingFor(
-        s.getProperties()
-    )
+    out_path = await s.getProperty("out_path")
+
+    return branch_config.do_register_gcroot(
+        await s.getProperty("default_branch"), await s.getProperty("branch")
+    ) and await s.getProperty("event") == "push" \
```
steveej, line 987: regarding https://github.com/nix-community/buildbot-nix/pull/312#issuecomment-2405883528 i expect this to prevent registering gcroots for PRs unless the source branch of the PR is one of the branches that match the regex and has set `register_gcroots = true`.

MagicRB: We cannot match on source branches sadly

MagicRB: Without making a network requests to girhubs api, or maybe, hm ill check it again, there might be a way. Maybe it arrives over the webhook but buildbot discards it?

steveej: i meant that if this branch that receives the push happens be a source branch for a PR then it register a GC root. under all other circumstances i foresee this to not register GC rootfs for PRs.

MagicRB: The pr branch names are `pull/number` or smth like that

steveej: got it, then i think your concern from the linked comment doesn't apply because the branch name won't ever match in case of a PR.

buildbot_nix/__init__.py
```diff
@@ -964,24 +974,26 @@ def nix_eval_config(
...
+      and not Path(str(gc_root)).exists() and Path(str(gc_root)).readlink() == str(out_path)
```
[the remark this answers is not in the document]
MagicRB: It will throw yes, but its fine ish, since we control that unless the instance admin goes out of their way to screw with it, itll be *fine*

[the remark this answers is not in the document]
steveej: > unless the instance admin goes out of their way to screw with it

hehe, poor admins take the blame :-) fine for now.

please do consider the code suggestion, it's unrelated to the exception discussion and important to fix IMO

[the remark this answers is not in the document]
MagicRB: Ah, right precedence, will apply thanks!

nix/master.nix
```diff
@@ -492,6 +492,75 @@ in
...
+                 registerGCRroots = true;
+                 copyOutputs = false;
+               }
+             }
+          ```
+          This example will result in `build_prs` and `register_gcroots` both being considered `true`,
+          but `copy_outputs` being `false` for the branch `foo`.
+
+          The default branches of all repos are considered to be matching of a rule setting all the options
+          to `true`.
```
Mic92, line 547:
```suggestion
An attribute set of branch rules specifies how to apply settings to specific branches using the `matchGlob` option. Each rule defines which branches it applies to, and the corresponding settings are applied to all branches that match the pattern. If multiple rules apply to a given branch, the rules are combined by "or"-ing each individual boolean option across all matching rules. For example:

```
{
  rule1 = {
    matchGlob = "f*";
    registerGCRroots = false;
    copyOutputs = false;
  }
  rule2 = {
    matchGlob = "foo";
    registerGCRroots = true;
    copyOutputs = false;
  }
}
```

In this example, for the branch `foo`, both `build_prs` and `register_gcroots` are considered `true`, while `copy_outputs` remains `false`.

By default, the main branches of all repositories match a rule that sets all options to `true`.
```

Mic92: https://gist.github.com/Mic92/9bdbd751ac1eb8200c8d5570aac0f936

MagicRB:
Holy... maybe I should squash these commits? 16 commits is a bit excessive...

PR #323, opened 2024-10-14
buildbot 4.1.0

zowoq:
https://nixpkgs-tracker.ocfox.me/?pr=348410

### DIFF
--- a/buildbot_nix/failed_builds.py
+++ b/buildbot_nix/failed_builds.py
@@ -22,7 +22,7 @@ class FailedBuild(BaseModel):
 
 class FailedBuildDB:
     def __init__(self, db_path: Path) -> None:
-        self.database = dbm.open(str(db_path), "c")
+        self.database = dbm.open(str(db_path), "c")  # noqa: SIM115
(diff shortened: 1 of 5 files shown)


flake.nix
```diff
@@ -3,7 +3,7 @@
   description = "A nixos module to make buildbot a proper Nix-CI.";
 
   inputs = {
-    nixpkgs.url = "github:Nixos/nixpkgs/nixos-unstable-small";
+    nixpkgs.url = "github:qowoz/nixpkgs/buildbot410";
```
Mic92, line 6: With this we should be able to add `buildbot` to nix/treefmt/flake-module.nix
and also remove:

```
[[tool.mypy.overrides]]
module = "buildbot.*"
ignore_missing_imports = true
```

from pyproject.toml.

Mic92: But this requires py.typed in the buildbot module.

Mic92 (approved):
Change looks good to me. @zowoq just merge when the PR hits the nixos channels.

PR #318, opened 2024-10-09
Per repo configuration

MagicRB:
This PR implements per repository configuration, it's currently quite minimal and only allows for overwriting:
1. the attribute (`--flake .#something`)
2. the reference lockfile (`--reference-lock-file somefile.lock`) provided that [329](https://github.com/nix-community/nix-eval-jobs/pull/329) is merged and a new enough `nix-eval-jobs` is present
3. #312 could be reimplemented on top of this, where in upon seeing `branches` inside a default branch evaluation `buildbot-nix` will save the regex into a file and use that when deciding which branches to consider

PR #325, opened 2024-10-30
reduce number of skipped build workers and pad worker name

zowoq:
Creating a skipped build worker for each thread is probably excessive, I think anything between 0.2 - 0.5 would be fine.

### DIFF
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -62,7 +62,9 @@ from .nix_status_generator import CombinedBuildEvent
 from .oauth2_proxy_auth import OAuth2ProxyAuth
 from .projects import GitBackend, GitProject
 
-SKIPPED_BUILDER_NAMES = [f"skipped-builds-{n}" for n in range(cpu_count())]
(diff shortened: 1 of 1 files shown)


buildbot_nix/__init__.py
```diff
@@ -62,7 +62,9 @@
 from .oauth2_proxy_auth import OAuth2ProxyAuth
 from .projects import GitBackend, GitProject
 
-SKIPPED_BUILDER_NAMES = [f"skipped-builds-{n}" for n in range(cpu_count())]
+SKIPPED_BUILDER_NAMES = [
+    f"skipped-builds-{n:03}" for n in range(int(cpu_count() * 0.25))
```
Mic92, line 66: Since it's partially io-bound, we can probably have a few more workers than cores.
Also we shouldn't go down to 0 if there someone has 1-3 cores.

```suggestion
    f"skipped-builds-{n:03}" for n in range(int(max(4, int(cpu_count() * 0.25))))
```

Mic92:
Tested and merged in https://github.com/nix-community/buildbot-nix/pull/318

PR #331, opened 2024-11-09
Actually enable `mypy`

MagicRB:
- [ ] some functions [1](https://github.com/nix-community/buildbot-nix/pull/331/files#diff-17cd44102ed0e7933f2157f30179888048c3beb5652a8fc4997f60183fedc427R465) [2](https://github.com/nix-community/buildbot-nix/pull/331/files#diff-17cd44102ed0e7933f2157f30179888048c3beb5652a8fc4997f60183fedc427L371) cannot be types as they don't fit into the subclass hierarchy, this is unfixable without changing the whole class hierarchy (buildbot likes this one a lot)
  ```python
  class A:
    def foo(a: int) -> int:
      return a
  class B:
    def foo(b: int, a: int) -> int: # this line will throw a type error
      return b + super().foo(a)
  ```
- [ ] `twisted.application.service.Service` [1](https://github.com/nix-community/buildbot-nix/pull/331/files#diff-17cd44102ed0e7933f2157f30179888048c3beb5652a8fc4997f60183fedc427L368) [2](https://github.com/nix-community/buildbot-nix/pull/331/files#diff-17cd44102ed0e7933f2157f30179888048c3beb5652a8fc4997f60183fedc427L461) doesn't type hint the `name` field, `mypy` assumes it has the type `None` and throws an error in `GitHubLegacySecretService` and `GitHubAppSecretService`
- [ ] `zope.interface` [1](https://github.com/nix-community/buildbot-nix/pull/331/files#diff-23f6eebd3059eac8e494796cdb5123595ff9c690e25518c08270ac48508acefbL19) doesn't have a py.typed marker and therefore will throw errors like `Skipping analyzing "zope.interface": module is installed, but missing library stubs or py.typed marker`
- [ ] `BuildNixEvalStatusGenerator.start_formatter` [1](https://github.com/nix-community/buildbot-nix/pull/331/files#diff-23f6eebd3059eac8e494796cdb5123595ff9c690e25518c08270ac48508acefbL105) and `BuildNixEvalStatusGenerator.end_formatter` were set to `IRenderable`, but `MessageFormatterBase` neither `MessageFormatterRenderable` is `IRenderable` (it doesn't implement the required signature either) so not sure what the type is supposed to be here

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 result
+result-*
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -29,7 +29,7 @@ from buildbot.process.project import Project
 from buildbot.process.properties import Properties
 from buildbot.process.results import ALL_RESULTS, statusToString, worst_status
 from buildbot.reporters.utils import getURLForBuildrequest
-from buildbot.schedulers.base import BaseScheduler
+from buildbot.schedulers.triggerable import Triggerable
 from buildbot.secrets.providers.file import SecretInAFile
 from buildbot.steps.trigger import Trigger
 from buildbot.www.authz import Authz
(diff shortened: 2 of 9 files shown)


MagicRB:
I'll fix the checks tomorrow

MagicRB:
Done, none of the things i detailed in the main PR comment have been fixed, i just slapped `# type: ignore [something]` on it which works for now.

PR #333, opened 2024-11-13
Pull based repos

MagicRB:


MagicRB:
@Mic92 im stuck, need you expertise, seems like the `GitPoller` is not triggering the schedulers no matter what I do, see [here](https://buildbot.redalder.org/#/projects/9), I also tried the following patch to hopefully catch the problem but it never actually run `filter_change`
```diff
diff --git a/buildbot_nix/__init__.py b/buildbot_nix/__init__.py
index c94f4a1..fa31e27 100644
--- a/buildbot_nix/__init__.py
+++ b/buildbot_nix/__init__.py
@@ -1261,14 +1261,25 @@ def config_for_project(
     failed_builds_db: FailedBuildDB,
     outputs_path: Path | None = None,
 ) -> None:
+    def foo(c: Any) -> bool:
+        print(c)
+        return c.branch == project.default_branch
+
+    class VerboseChangeFilter(util.ChangeFilter):
+        def filter_change(self, change: Any):
+            print(change)
+            ret = super().filter_change(change)
+            print(ret)
+            return ret
+
     config["projects"].append(Project(project.name))
     config["schedulers"].extend(
         [
             schedulers.SingleBranchScheduler(
                 name=f"{project.project_id}-default-branch",
-                change_filter=util.ChangeFilter(
+                change_filter=VerboseChangeFilter(
                     repository=project.url,
-                    filter_fn=lambda c: c.branch == project.default_branch,
+                    filter_fn=lambda c: foo(c),
                 ),
                 builderNames=[f"{project.name}/nix-eval"],
                 treeStableTimer=5,
```

nix/master.nix
```diff
@@ -444,6 +444,99 @@ in
...
+            number of seconds. Can be used to avoid a thundering herd situation.
+          '';
+        };
+
+        sshPrivateKeyFile = lib.mkOption {
+          type = lib.types.nullOr lib.types.path;
+          default = null;
+          description = ''
+            If non-null the specified SSH key will be used to fetch all configured repositories.
+            This option is overriden by the per-repository `sshPrivateKeyFile` option.
```
Mic92, line 525: The other option seems to be not actually implemented. So we either remove it from the description or implement it.

Mic92:
Works!

PR #342, opened 2024-12-01
buildbot_effects: set uid, gid, bind ssh known hosts

zowoq:


buildbot_effects/__init__.py
```diff
@@ -222,13 +222,20 @@ def run_effects(
         "/proc",
         "--dev",
         "/dev",
+        "--uid",
+        "0",
+        "--gid",
+        "0",
```
Mic92, line 228: What is the default? Should we not just keep the uid rather than pretending to be root? Some applications might than behave funny.

buildbot_effects/__init__.py
```diff
@@ -222,13 +222,20 @@ def run_effects(
...
         "--ro-bind",
         "/etc/resolv.conf",
         "/etc/resolv.conf",
         "--ro-bind",
         "/etc/hosts",
         "/etc/hosts",
         "--ro-bind",
+        "/etc/ssh/ssh_known_hosts",
+        "/etc/ssh/ssh_known_hosts",
```
Mic92, line 237: That seems ok with me. But to be compatible with hercules-ci-effects, what's the default there?

buildbot_effects/__init__.py
zowoq: Default seems to be root:

https://github.com/hercules-ci/hercules-ci-effects/blob/b89ac4d66d618b915b1f0a408e2775fe3821d141/effects/effect/effect.nix#L79

buildbot_effects/__init__.py
zowoq: Hercules examples show writing to `~/.ssh`.

I wanted to reuse the programs.ssh.knownHosts config that is already on disk rather than needing to evaluate it in each effect.

Mic92: > Hercules examples show writing to `~/.ssh`.
> 
> I wanted to reuse the programs.ssh.knownHosts config that is already on disk rather than needing to evaluate it in each effect.

Could you not just import the file and than use the example: https://github.com/hercules-ci/hercules-ci-effects/blob/b89ac4d66d618b915b1f0a408e2775fe3821d141/docs/modules/ROOT/pages/guide/deploy-a-nixos-machine.adoc#L70

Than it also works locally outside of CI.

zowoq: I think mounting /etc/ssh/ssh_known_hosts is simpler than the alternatives, I don't understand the objection?

zowoq: > Only issue with that is that would allow nix-community member to bind mount arbitrary files from the filesystem? That might be also an issue with our existing hercules-ci installation.

No, mounts would need to be added to the hercules config in infra first before they can be used in effects.

Mic92: > I think mounting /etc/ssh/ssh_known_hosts is simpler than the alternatives, I don't understand the objection?
> 
> Why is /etc/hosts okay and not this?

Because the original implementation doesn't implement it like this and it also would make ci results depending more on the host configuration, i.e. the buildbot-effect ci can be also run locally but would than fail to run because it's missing the host keys that the CI servers have.

zowoq: So /etc/hosts should be removed instead?

zowoq: > `/etc/hosts` is also mounted by the hercules effects api and it's needed to provide resolving `localhost`. If I had implemented, I would have chosen a more minimal hosts file, but this mirrors the hercules-ci effects implementation.

I can't find this, can you link to it please?

Mic92: You are right. Hercules-ci also doesn't write `/etc/hosts`. Dropping it in https://github.com/nix-community/buildbot-nix/pull/343

buildbot_effects/__init__.py
Mic92: Ok than.

PR #345, opened 2024-12-07
OAuth2ProxyAuth: initializing AuthBase.userInfoProvider

ajamtli:
Setting up buildbot-nix with oauth2-proxy gives error:

```
buildbot_nix/oauth2_proxy_auth.py", line 25, in __init__
twistd[708]: if self.user_info_provider is None:
twistd[708]:     builtins.AttributeError: 'OAuth2ProxyAuth' object has no attribute 'user_info_provider'

```
There is also an issue that the user isn't granted 'admins' role if user_info_provider is used instead of userInfoProvider.


MagicRB:
LGTM, good catch, thanks!

MagicRB (commented):
.

PR #346, opened 2024-12-09
nix/worker: update trace for buildbot-nix.packages

zowoq:
trace: Obsolete option `services.buildbot-nix.worker.package' is used ...


Mic92:
@mergify queue

PR #362, opened 2024-12-29
Support using SSH to fetch repositories on Gitea

MagicRB:
I'll extend this to to GitHub before merging, maybe also unify the ssh private key options between the backends


Mic92:
Done in https://github.com/nix-community/buildbot-nix/pull/413

PR #366, opened 2025-01-01
Return a result (`int`) from `BuildTrigger.run` as BuildBot expects

MagicRB:


MagicRB:
Closed in favor of #367

PR #367, opened 2025-01-01
add missing return

Mic92:
this was introduced in f5f907d0


MagicRB:
Works for me

Mic92:
@mergify queue

PR #372, opened 2025-01-07
Prepare color support

Mic92:


Mic92:
This requires https://github.com/NixOS/nix/pull/12148 to take full effect. But it will work as is.

PR #382, opened 2025-01-21
buildbot_effects: bind nix daemon socket, set NIX_REMOTE

zowoq:
https://github.com/hercules-ci/hercules-ci-agent/blob/cae08186393d9736f2b3a5d30b2c7efe5569e337/hercules-ci-agent/src/Hercules/Effect.hs

The socket handling in hercules isn't simple, not sure if any more of it is necessary here.


Mic92:
It seems there is some proxy that allows to inject additional nix options? I don't see a usecase for this yet.

Mic92:
@mergify queue

PR #384, opened 2025-01-23
 buildbot-effects: don't print program arguments by default

Mic92:


Mic92:
@mergify refresh

Mic92:
@mergify requeue

Mic92:
@mergify requeue

PR #388, opened 2025-02-04
Skip builds fully to avoid delays from worker setup

MagicRB:
Support skipped builds, depends on https://github.com/buildbot/buildbot/pull/8288 which won't apply to our version of buildbot. I'll try a backport


Mic92:
https://github.com/nix-community/buildbot-nix/pull/419

PR #389, opened 2025-02-04
Report `nix-eval` as failed if effect fails to run

MagicRB:


zowoq:
I missed this PR when it was opened, I see the commit was merged in https://github.com/nix-community/buildbot-nix/pull/392.

It doesn't seem to make a difference, effect fails but the github check is green.

MagicRB:
@zowoq which check precisely?

Mic92:
@MagicRB he says that if an effect fails the CI is green. We actually do not emit any github status for effects at all, which is the actual issue.

MagicRB:
The `nix-eval` should fail if an effect fails afaik.

Mic92:
Did we not change that when added `nix-build` github statuses?

MagicRB:
Ah hell, youre right, the build fails in buidlbot but not on the report side because we dont report it normally. Ill make a new PR, should i fold it into `nix-eval` or make a new report completely, as in `buildbot-effects`?

Mic92:
> Ah hell, youre right, the build fails in buidlbot but not on the report side because we dont report it normally. Ill make a new PR, should i fold it into `nix-eval` or make a new report completely, as in `buildbot-effects`?

Make a new one.

MagicRB:
Will do

PR #392, opened 2025-02-11
nix-eval-jobs: fix gcroots dir not correctly passed

Mic92:


MagicRB:
ah so that's why my buildbot gcroots dir contains a lot of `Insecure` lmao

PR #394, opened 2025-02-12
Add `users` option to both GitHub and Gitea restricting project owners

### LINKED ISSUE #393: Add way to ignore installations based on the owner of the repository
Say you're running a GitHub App and want to run it in multiple organizations, currently there is no way to restrict it to just 2 organizations but not *any* organization. GitHub doesn't provide this level of control over applications. A similar problem exists on Forgejo, if you define create a buildbot user, any repository which qualifies (i.e. has a matching topic and the buildbot user is an admin for it) will be automatically loaded as a project. This isn't a problem for private or restricted signup Forgejo instances, but for public ones like `https://codeberg.org` one may end up running other people's CI, which in the best case means an increased electricity bill, or in the worst case a compromised system.

MagicRB:
Hopefully closes #393


MagicRB:
@zowoq this looks good? I tested the Gitea backend at it works, didn't test the GitHub one as I don't have GitHub setup, but it should work. @Mic92 can you test?

zowoq:
Tested removing the topic and using only the repoAllowlist in nix-community, adding/removing repos works, did need to change the `buildbot-nix` topic default from `build-with-buildbot` to `null`.

```diff
-      topic = "nix-community-buildbot";
+      repoAllowlist = [
+        "nix-community/authentik-nix"
+        "nix-community/autofirma-nix"
+        "nix-community/dream2nix"
+        "nix-community/ethereum.nix"
+        "nix-community/infra"
+        "nix-community/lanzaboote"
+        "nix-community/neovim-nightly-overlay"
+        "nix-community/nix-direnv"
+        "nix-community/nix-index"
+        "nix-community/nix4nvchad"
+        "nix-community/NixNG"
+        "nix-community/nixos-facter"
+        "nix-community/nixos-facter-modules"
+        "nix-community/nixos-generators"
+        "nix-community/nixos-images"
+        "nix-community/nixpkgs-update"
+        "nix-community/nixpkgs-xr"
+        "nix-community/nixvim"
+        "nix-community/raspberry-pi-nix"
+        "nix-community/srvos"
+      ];
```

```diff
diff --git a/nix/master.nix b/nix/master.nix
index 2a75c8f74..3a3fc8854 100644
--- a/nix/master.nix
+++ b/nix/master.nix
@@ -374,7 +374,7 @@ in
         };
         topic = lib.mkOption {
           type = lib.types.nullOr lib.types.str;
-          default = "build-with-buildbot";
+          default = null;
           description = ''
             Projects that have this topic will be built by buildbot.
             If null, all projects that the buildbot Gitea user has access to, are built.
@@ -453,7 +453,7 @@ in
         };
         topic = lib.mkOption {
           type = lib.types.nullOr lib.types.str;
-          default = "build-with-buildbot";
+          default = null;
           description = ''
             Projects that have this topic will be built by buildbot.
             If null, all projects that the buildbot github user has access to, are built.
```

MagicRB:
You should be able to just set it to null no? I think that building *everything* by default is not a good idea.

Mic92:
> You should be able to just set it to null no? I think that building _everything_ by default is not a good idea.

I think this was just for testing not intended as the new default.

nix/master.nix
```diff
@@ -337,6 +337,18 @@ in
           default = cfg.authBackend == "gitea";
         };
 
+        userAllowlist = lib.mkOption {
+          type = lib.types.nullOr (lib.types.listOf lib.types.str);
+          default = null;
+          description = "If non-null, specifies users/organizations that are allowed to use buildbot, i.e. buildbot-nix will ignore any repositories not owned by these users/organizations.";
```
Mic92, line 343: Should we mention somewhere that this is on top of the buildbot topic?

zowoq:
> > You should be able to just set it to null no? I think that building _everything_ by default is not a good idea.
> 
> I think this was just for testing not intended as the new default.

Was kind of thinking of a new default. Could have an assert requiring that one or more of topic, userAllowlist, repoAllowlist is set? But maybe not be worth the disruption for other users.

zowoq:
Can we move forward with this?

zowoq:
I've rebased to fix the merge conflict, are we okay with merging?

Mic92:
@zowoq please merge if you have tested it in nix-community.

nixosModules/master.nix
```diff
@@ -336,6 +348,18 @@ in
           default = cfg.authBackend == "github";
         };
 
+        userAllowlist = lib.mkOption {
+          type = lib.types.nullOr (lib.types.listOf lib.types.str);
+          default = null;
+          description = "If non-null, specifies users/organizations that are allowed to use buildbot, i.e. buildbot-nix will ignore any repositories not owned by these users/organizations.";
+        };
+
+        repoAllowlist = lib.mkOption {
```
Mic92, line 357: Would be cool if you could add those options also to https://github.com/nix-community/buildbot-nix/blob/main/examples/master.nix
To have same example-driven documentation.

zowoq:
`reload_github_projects` fails when an a repo / github installation is removed, seems the various buildbot github app / project json files that are cached aren't regenerated so it fails trying to access the removed repo / installation.
```
Failed to reload project list: Traceback (most recent call last):
  File "/nix/store/hyrcaa8b7dqigaaailf3hh6ixfsp0hnj-python3-3.12.10-env/lib/python3.12/site-packages/twisted/python/threadpool.py", line 269, in inContext
    result = inContext.theWork()  # type: ignore[attr-defined]
  File "/nix/store/hyrcaa8b7dqigaaailf3hh6ixfsp0hnj-python3-3.12.10-env/lib/python3.12/site-packages/twisted/python/threadpool.py", line 285, in <lambda>
    inContext.theWork = lambda: context.call(  # type: ignore[attr-defined]
  File "/nix/store/hyrcaa8b7dqigaaailf3hh6ixfsp0hnj-python3-3.12.10-env/lib/python3.12/site-packages/twisted/python/context.py", line 117, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/nix/store/hyrcaa8b7dqigaaailf3hh6ixfsp0hnj-python3-3.12.10-env/lib/python3.12/site-packages/twisted/python/context.py", line 82, in callWithContext
    return func(*args, **kw)
  File "/nix/store/hyrcaa8b7dqigaaailf3hh6ixfsp0hnj-python3-3.12.10-env/lib/python3.12/site-packages/buildbot_nix/github_projects.py", line 193, in run_deferred
    v.get(),
  File "/nix/store/hyrcaa8b7dqigaaailf3hh6ixfsp0hnj-python3-3.12.10-env/lib/python3.12/site-packages/buildbot_nix/github/installation_token.py", line 99, in get
    self.verify()
  File "/nix/store/hyrcaa8b7dqigaaailf3hh6ixfsp0hnj-python3-3.12.10-env/lib/python3.12/site-packages/buildbot_nix/github/installation_token.py", line 107, in verify
    self.token, self.expiration = InstallationToken._generate_token(
  File "/nix/store/hyrcaa8b7dqigaaailf3hh6ixfsp0hnj-python3-3.12.10-env/lib/python3.12/site-packages/buildbot_nix/github/installation_token.py", line 46, in _generate_token
    token = InstallationToken._create_installation_access_token(
  File "/nix/store/hyrcaa8b7dqigaaailf3hh6ixfsp0hnj-python3-3.12.10-env/lib/python3.12/site-packages/buildbot_nix/github/installation_token.py", line 35, in _create_installation_access_token
    return http_request(
  File "/nix/store/hyrcaa8b7dqigaaailf3hh6ixfsp0hnj-python3-3.12.10-env/lib/python3.12/site-packages/buildbot_nix/common.py", line 98, in http_request
    raise HttpError(msg) from e
buildbot_nix.common.HttpError: Request for POST https://api.github.com/app/installations/70038369/access_tokens failed with 404 Not Found: {"message":"Not Found","documentation_url":"https://docs.github.com/rest/reference/apps#create-an-installation-access-token-for-an-app","status":"404"}
```

zowoq:
@Mic92 I think this problem is the same as the pre-existing issue where reload project sometimes doesn't take effect, seems to be something wrong with the caching. As this PR doesn't change the default and I don't think it'll make anything worse I'd like to merge this.

Mic92:
Okay

PR #400, opened 2025-02-20
Rewrite cachix config as pure nix

antifuchs:
This PR untangles configuration of the cachix cache slightly, removing the special python code for it and setting the equivalent configuration all up in nix. Also, it uses the opportunity to group all cachix-related configuration into a single mkMerged block in master.nix and refactors the flake's `lib` export such that it can be used inside master.nix.

### DIFF
--- a/buildbot_nix/buildbot_nix/__init__.py
+++ b/buildbot_nix/buildbot_nix/__init__.py
@@ -1791,20 +1791,6 @@ class NixConfigurator(ConfiguratorBase):
 
         eval_lock = util.MasterLock("nix-eval")
 
-        if self.config.cachix is not None:
-            self.config.post_build_steps.append(
-                models.PostBuildStep(
-                    name="Upload cachix",
-                    environment=self.config.cachix.environment,
-                    command=[
-                        "cachix",
-                        "push",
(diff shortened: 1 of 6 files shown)


MagicRB:
@antifuchs PR looks good, factored out the cachix stuff into a separate file, just to shrink `master.nix`, it's been too big for quite some time in my opinion.

MagicRB:
I assume you've tested this to work, I don't have cachix hooked up on my instances.

antifuchs:
> I assume you've tested this to work, I don't have cachix hooked up on my instances.

Testing it, looks like something is wrong now - it says it can't find the signing key, which might indicate that we mangled the CLI invocation. Debugging now...

antifuchs:
(Gentle ping on this, I tested this and it's working as it was before)

Mic92:
Justing testing this feature. Don't take it seriously

Mic92:
Ok. Also not very useful from the looks of it.

Mic92:
@zowoq would you be able to test this change?

zowoq:
> would you be able to test this change?

Deployed it on the community buildbot for a bit, seems fine, cachix works.

PR #402, opened 2025-02-25
Support translating from "repo_name" in GitHub secret provider

MagicRB:
With this the following works `Secret("github-token-my/repo")`, no longer requiring the knowledge of the particular installation ID. This is particularly handy if you want to rely on `buildbot-nix` GitHub App integration from another buildbot plugin. It's cursed but it works.


Mic92:
I don't quite understand the usecase. What is the other buildbot plugin, you want to use?

MagicRB:
@Mic92 https://git.numtide.com/MagicRB/buildbot-renovate there it is

PR #410, opened 2025-03-27
worker: remove pinned nix-eval-jobs

zowoq:
https://nixpk.gs/pr-tracker.html?pr=393621


nix/worker.nix
```diff
@@ -50,11 +62,6 @@ in
         default = config.networking.hostName;
         description = "The buildbot worker name.";
       };
-      nixEvalJobs.package = lib.mkOption {
-        type = lib.types.package;
-        default = pkgs.callPackage ./nix-eval-jobs.nix { };
```
zimbatm, line 55: Might be best to keep this option and change the default to `pkgs.nix-eval-jobs`. It will make it eaiser to test new nix-eval-jobs versions.

Mic92: Should be fine.

PR #416, opened 2025-04-10
buildbot_effects: bind /bin/sh

zowoq:
Error started with 2.24 -> 2.28 nix bump. https://github.com/nix-community/infra/pull/1768#issuecomment-2788784442

Looks like it is from https://github.com/NixOS/nix/commit/a03bb4455cee010bbfcf7e322b10ec7e35123032.
```
Couldn't execute /bin/sh -c "echo started": No such file or directory
```

### DIFF
--- a/buildbot_effects/buildbot_effects/__init__.py
+++ b/buildbot_effects/buildbot_effects/__init__.py
@@ -240,6 +240,9 @@ def run_effects(
         "--bind",
         "/nix/var/nix/daemon-socket/socket",
         "/nix/var/nix/daemon-socket/socket",
+        "--ro-bind",
+        "/bin/sh",
+        "/bin/sh",
     ]
 
     with NamedTemporaryFile() as tmp:
--- a/buildbot_nix/buildbot_nix/worker.py
+++ b/buildbot_nix/buildbot_nix/worker.py
@@ -25,11 +25,12 @@ class WorkerConfig:
     worker_name: str = field(
         default_factory=lambda: os.environ.get("WORKER_NAME", socket.gethostname())
     )
-    worker_count_str: str | None = os.environ.get("WORKER_COUNT")
-    if worker_count_str is not None:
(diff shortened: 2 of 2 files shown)


buildbot_effects/buildbot_effects/__init__.py
```diff
@@ -240,6 +240,9 @@ def run_effects(
         "--bind",
         "/nix/var/nix/daemon-socket/socket",
         "/nix/var/nix/daemon-socket/socket",
+        "--ro-bind",
```
Mic92, line 243: @roberth you might also want to do this for hercules ci effects?

roberth:
I believe `/bin/sh` in the Linux Nix sandbox was a mistake, and we don't have to repeat it for effects.

The old behavior was [intentional](https://docs.hercules-ci.com/hercules-ci-agent/effects#paths), and this should be fixed in `hercules-ci-effects` instead.

> The `/bin/sh` and `/usr/bin/env` executables are _not_ provided by the sandbox and should be installed by the effect instead.

Mic92:
> I believe `/bin/sh` in the Linux Nix sandbox was a mistake, and we don't have to repeat it for effects.
> 
> The old behavior was [intentional](https://docs.hercules-ci.com/hercules-ci-agent/effects#paths), and this should be fixed in `hercules-ci-effects` instead.
> 
> > The `/bin/sh` and `/usr/bin/env` executables are _not_ provided by the sandbox and should be installed by the effect instead.

@roberth I am not sure we have really a chance in leaving out /bin/sh. This would fundamentally break [system()](https://man7.org/linux/man-pages/man3/system.3.html) in glibc.

Mic92:
I agree we don't need /usr/bin/env

roberth:
I don't think we need to leave it out. My main issue is that `/bin/sh` was impure.
- https://github.com/hercules-ci/hercules-ci-effects/pull/187

PR #420, opened 2025-04-16
Add gitlab integration

### LINKED ISSUE #329: Gitlab integration?
This was asked during NixCon, but I figure I'll lift it here in order to track progress: is Gitlab in-scope, and if so, what work remains?

wlcx:
This is a draft currently as there's some tyre kicking still to do, but I thought it was worth putting up for some initial thoughts!

resolves #329 

Still to do:
- [ ] avatars
- [ ] why don't builds get triggered on commit?
- [ ] why aren't statuses working?
- [ ] probably other stuff

### DIFF
--- a/buildbot_nix/buildbot_nix/gitlab_project.py
+++ b/buildbot_nix/buildbot_nix/gitlab_project.py
@@ -1,0 +1,319 @@
+import os
+import signal
+from pathlib import Path
+from typing import Any
+from urllib.parse import urlparse
+
+from buildbot.changes.base import ChangeSource
+from buildbot.config.builder import BuilderConfig
+from buildbot.plugins import util
+from buildbot.reporters.base import ReporterBase
+from buildbot.reporters.gitlab import GitLabStatusPush
+from buildbot.www.auth import AuthBase
+from buildbot.www.avatar import AvatarBase
+from pydantic import BaseModel
+from twisted.logger import Logger
+from twisted.python import log
+
+from buildbot_nix.common import (
+    ThreadDeferredBuildStep,
+    atomic_write_file,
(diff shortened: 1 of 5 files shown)


Mic92:
Looks ok so far.

buildbot_nix/buildbot_nix/gitlab_project.py
```diff
@@ -0,0 +1,317 @@
...
+            lambda repo: repo.topics,
+        )
+        atomic_write_file(self.project_cache_file, model_dump_project_cache(repos))
+
+    def run_post(self) -> Any:
+        return util.SUCCESS
+
+
+class CreateGitlabProjectHooks(ThreadDeferredBuildStep):
+    name = "create_gitlab_project_hooks"
```
Mic92, line 237: Btw. if gitlab has a different way to install webhooks i.e. through some app interface, this might be preferred. It's a bit of a historical accident in github that we didn't just use the webhooks that are provided by a github app rather than installing an explicit webhook into each repository.

greg-hellings:
Ooo, I'm eager to have this. I've been building all my Nix targets using a dynamic fan-out Pipeline in Gitlab CI right now. It's working OK, but I know there are significant efficiencies that this Buildbot config uses and I can't wait to taste them.

Mic92:
@greg-hellings maybe you can team up with @wlcx
We have a matrix channel at #buildbot-nix:thalheim.io

greg-hellings:
@Mic92 Happily! I can definitely look at the list of things he has above still TBD. I'm in the midst of migrating hypervisors, so you won't see me back on Matrix over this weekend. I should be there Monday to collaborate on this.

PR #430, opened 2025-04-29
Add a local dev workflow

Mic92:


packages/master.cfg.py
```diff
@@ -0,0 +1,55 @@
...
+
+STATE_DIR = Path(".")
+PORT = 8012
+
+buildbot_nix_config = BuildbotNixConfig(
+    db_url="sqlite:///state.sqlite",
+    auth_backend="none",
+    gitea=None,
+    github=None,
+    pull_based=dict(
```
Mic92, line 20: There is room for improvements in this file, but I kind of want to have some basic thing working because the current iteration cycles are kind of long.

PR #433, opened 2025-05-03
Add lib.interpolate function back

antifuchs:
This seems to have gotten dropped in #429, but it's part of the exported interface by now (and I use & need it for accessing my attic cache).

This PR adds back the `lib` flake attribute, and the `interpolate` attribute; this time it includes documentation (in nix-community/nixdoc format).

### DIFF
--- a/nix/lib.nix
+++ b/nix/lib.nix
@@ -1,0 +1,24 @@
+{
+  /**
+    Marks a string as a value that will be interpolated on the buildbot master or worker.
+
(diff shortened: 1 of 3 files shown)


nix/lib.nix
```diff
@@ -0,0 +1,24 @@
...
+    interpolate :: String -> InterpolateString
+    ```
+
+    # Arguments
+
+    value
+    : The interpolation format string.
+  */
+  interpolate = value: {
+    _type = "interpolate";
```
Mic92, line 21: Why does this need a separate type?

PR #438, opened 2025-05-22
Add option to show stack trace in failed builds

Erethon:
This adds an option called `showTrace` that adds `--show-trace` on nix builds and evals to show the full stack trace when something fails. It's off by default to keep the existing functionality as is.

You can see it in action [here](https://buildbot.ngi.nixos.org/#/builders/175/builds/1800).

I'm not that familiar with the buildbot(-nix) codebase, so I hope I haven't missed something obvious in this change.


Erethon:
> Needs a `nix fmt` and `git commit --amend` locally to fix up the formatting.

Sorry about that, should have opened as draft until all the checks passed.

PR #472, opened 2025-08-27
nix-eval-jobs: disable eval-cache

Mic92:


zowoq:
Should this be done for hydra as well?

Mic92:
> Should this be done for hydra as well?

I think it would make sense - should reduce lock contention which is even more important when dealing with many nixpkgs as hydra does.

PR #494, opened 2025-09-17
Various refactorings

Mic92:


buildbot_nix/buildbot_nix/nix_eval.py
```diff
@@ -0,0 +1,639 @@
...
+            logEnviron=False,
+        ),
+    )
+    factory.addStep(
+        BuildbotEffectsCommand(
+            project=project,
+            env={},
+            name="Evaluate effects",
+            command=[
+                "buildbot-effects",
```
zowoq, line 616: https://buildbot.thalheim.io/#/builders/105/builds/302/steps/6/logs/stdio

I noticed that repo doesn't seem work (not actually sure if it ever worked). rev, branch, repo all seem to handled automatically by buildbot-effects, is there reasons to set them explicitly?

Mic92: Can you make an issue for this?

PR #513, opened 2025-09-26
fix: update project maps directly in reload services for GitHub and G…

Mic92:
…itea

When systemd removes cache files on startup, both GitHub and Gitea backends have stale project maps that cause status updates to be incorrectly skipped.

This fix updates the in-memory project maps directly when the reload services complete:
- GitHub: Updates project_id_map in ReloadGithubInstallations
- Gitea: Updates gitea_projects set in ReloadGiteaProjects

The SIGHUP signal will still trigger a full config reload for consistency, but the direct updates ensure the maps are current immediately after reload without waiting for the signal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures Gitea and GitHub project/repository lists refresh consistently after reloads, reducing stale or missing entries.
  * Keeps in-memory backend data synchronized with the latest reload state for more reliable behavior.

* **Chores**
  * Improved backend synchronization flow so project maps and caches stay up to date post-reload.
  * Enhanced logging around project reloads to aid monitoring and troubleshooting.

### DIFF
--- a/buildbot_nix/buildbot_nix/gitea_projects.py
+++ b/buildbot_nix/buildbot_nix/gitea_projects.py
@@ -181,7 +181,9 @@ class GiteaBackend(GitBackend):
         """Updates the flake an opens a PR for it."""
         factory = util.BuildFactory()
         factory.addStep(
-            ReloadGiteaProjects(self.config, self.config.project_cache_file),
+            ReloadGiteaProjects(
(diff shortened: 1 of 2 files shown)


Mic92:
@zowoq I just deleted the project map as described and redployed this fix on my server.
Without having to do anything it was able to load all my projects: https://github.com/Mic92/dotfiles/pull/3483

PR #557, opened 2026-01-20
`nix/clan-module`: init

MagicRB:
The new module options doc [rendered](https://github.com/MagicRB/buildbot-nix/blob/feat/clan-module/docs/OPTIONS.md)


MagicRB:
@Mic92 i added a super basic option doc thing in the last commit, with the split options the approach of "look at the source code" wont work. IMO it would be too hard for people to parse 2 separate Nix files, especially with the mapping i have to do. Not sure how we could have MD option docs for this repo, ive never done that. Any suggestions?

PR #614, opened 2026-04-17
Allow specifying extra sandbox paths for effect sandbox

MagicRB:


buildbot_nix/buildbot_nix/buildbot_effects.py
```diff
@@ -50,6 +58,14 @@ def buildbot_effects_config(
                     util.Property("branch"),
                     "--repo",
                     util.Property("project"),
+                    *functools.reduce(
```
Mic92: this seems slower and more code than just using list comprehensions.

PR #619, opened 2026-04-19
gitea: fix repos with correct topic not found

Enzime:
This is the same change as https://github.com/Mic92/gitea-mq/pull/23

Currently if `buildbot-nix` is a Gitea-wide administrator and a repo has the correct topic, the repo won't be discovered unless `buildbox-nix` is added as a collaborator

Marking it as a draft as I haven't tested this by hand yet

### DIFF
--- a/buildbot_nix/buildbot_nix/gitea_projects.py
+++ b/buildbot_nix/buildbot_nix/gitea_projects.py
@@ -415,9 +415,16 @@ class ReloadGiteaProjects(ThreadDeferredBuildStep):
 def refresh_projects(config: GiteaConfig) -> list[RepoData]:
     repos = []
 
+    # Use /repos/search so that site admins discover all repos on the instance,
+    # not just repos they personally own or collaborate on.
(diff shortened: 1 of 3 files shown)


Enzime:
I tested this by hand and it works 👍

MagicRB:
Does the old behavior still work?

buildbot_nix/buildbot_nix/gitea_projects.py
```diff
@@ -415,25 +415,23 @@ def run_post(self) -> Any:
 def refresh_projects(config: GiteaConfig) -> list[RepoData]:
     repos = []
 
+    # Use /repos/search so that site admins discover all repos on the instance,
+    # not just repos they personally own or collaborate on.
+    search_url = f"{config.instance_url}/api/v1/repos/search?limit=100"
```
Mic92, line 420: Is this desired behavior? What if you multiple orgs or registration open and start building random projects from other people just because you have access?

PR #629, opened 2026-05-05
Use merge queue scheduler for gitea-mq

Enzime:
Without this two Gitea merge queue branches created at the same time will get ignored due to the tree stable timer as part of the primary scheduler

Draft as untested :)


Enzime:
Tested this manually and it works 👍

